PR #37, opened 2019-10-11
metrics: Implement hack clubhouse metrics

danigm:
This patch implements the following metrics produced by the
com.hack_computer.Clubhouse app:

 * HackClubhouseChangePage
 * HackClubhouseEnterPathway
 * HackClubhouseAchievementPoints
 * HackClubhouseAchievement
 * HackClubhouseProgress

### DIFF
--- a/azafea/event_processors/metrics/events/__init__.py
+++ b/azafea/event_processors/metrics/events/__init__.py
@@ -13,7 +13,7 @@ from gi.repository import GLib
 
 from sqlalchemy.dialects.postgresql import DOUBLE_PRECISION, JSONB
 from sqlalchemy.schema import Column
-from sqlalchemy.types import ARRAY, BigInteger, Boolean, Integer, LargeBinary, Unicode
+from sqlalchemy.types import ARRAY, BigInteger, Boolean, Integer, LargeBinary, Numeric, Unicode
 
 from azafea.vendors import normalize_vendor
 from ..utils import get_asv_dict, get_bytes, get_strings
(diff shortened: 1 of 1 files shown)


azafea/event_processors/metrics/events/__init__.py
```diff
@@ -533,6 +575,22 @@ def _get_fields_from_payload(payload: GLib.Variant) -> Dict[str, Any]:
...
+
+    complete = Column(Boolean, nullable=False)
+    pathways = Column(ARRAY(Unicode, dimensions=1), nullable=False)
+
+    @staticmethod
+    def _get_fields_from_payload(payload: GLib.Variant) -> Dict[str, Any]:
+        return {
+            'complete': payload.get_child_value(0).get_boolean(),
+            'pathways': get_strings(payload.get_child_value(1)),
+        }
```
bochecha: As it is, only the start payload will be parsed. The stop payload will be completely discarded.

Are you ok with that?

If not, then I suggest we don't implement this event at all until Azafea grows the capacity to handle sequence events better.

danigm: Okay, I've removed the QUEST event completely so we can add in the future when it's fully supported.

azafea/event_processors/metrics/events/__init__.py
```diff
@@ -165,6 +165,79 @@ def _get_fields_from_payload(payload: GLib.Variant) -> Dict[str, Any]:
...
+    __payload_type__ = 'a{sv}'
+
+    complete = Column(Boolean, nullable=False)
+    quest = Column(Unicode, nullable=False)
+    pathways = Column(ARRAY(Unicode, dimensions=1), nullable=False)
+    progress = Column(Numeric, nullable=False)
+
+    @staticmethod
+    def _get_fields_from_payload(payload: GLib.Variant) -> Dict[str, Any]:
+        return get_asv_dict(payload)
```
bochecha: This will work as long as you don't remove any field from the payload, or change their types.

(I usually keep `get_asv_dict` for the cases where the columns aren't well defined, and we store the whole dict as a single `JSONB` field, because it is a bit fragile otherwise)

If you're ok with this caveat though, I see no reason to block this. :slightly_smiling_face:

danigm: It's okay, the payload shouldn't change, it's not a dynamic payload, it'll be the same for all events.

PR #70, opened 2019-12-23
Enable row-level security

bochecha:
This will allow setting things up on the PostgreSQL side so that most
accounts can only view the user data for the deployments they are
responsible for.

For example, the people at Endless Solutions with access to the database
will only see data for Solutions machines, not for all other users.

The actual security policies will be created and maintained in the
deployment configuration in Terraform, because the azafea user (owning
the database) doesn't have the permissions to do those.

### DIFF
--- a/azafea/event_processors/metrics/v2/migrations/27156a7be72d_enable_row_level_security.py
+++ b/azafea/event_processors/metrics/v2/migrations/27156a7be72d_enable_row_level_security.py
@@ -1,0 +1,25 @@
+# type: ignore
+
+"""Enable row-level security
+
+Revision ID: 27156a7be72d
+Revises: 7f4c0154d6cd
+Create Date: 2019-12-23 10:29:06.431725
+
+"""
+from alembic import op
+
+
+# revision identifiers, used by Alembic.
+revision = '27156a7be72d'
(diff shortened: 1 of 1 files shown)


bochecha:
> I'm not familiar with alembic, but this all looks fine to me. Where do the revisions come from?

They all come from Azafea, each event processor comes with its own migrations:

* activation: https://github.com/endlessm/azafea/tree/9a01f82e676c4f9fc7ade961bd38c50b7dfc0355/azafea/event_processors/activation/v1/migrations
* metrics: https://github.com/endlessm/azafea/tree/dbfe29ffaa12081d54f6af846489077198421fdb/azafea/event_processors/metrics/v2/migrations
* ping: https://github.com/endlessm/azafea/tree/16e0f559f1fba02f63e045e7568421920cc24271/azafea/event_processors/ping/v1/migrations

---

However, this won't be needed, @adarnimrod is enabling the row-level security in Terraform instead.

bochecha:
Confirmed with @adarnimrod that we're going a different way to deploy row-level security.

adarnimrod:
We can drop this PR, I handled it in Terraform using `psql` and `local-exec`.

PR #73, opened 2019-12-31
Embrace containers for the CI

bochecha:
Now that we can have Docker images for development, this also moves the CI to use it instead of whatever Travis does to their poor old Ubuntu 16.04.

Controlling our test environment with containers frees us from the underlying host, so we can use a more minimal Travis host image and upgrade to a newer Ubuntu for the host.

Finally, this fixes a minor problem with our CI where some parts of an undesired PostgreSQL version would remain installed. While that never caused any problem, removing it avoids all potential for any such problems in the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,10 +5,8 @@ branches:
   only:
   - master
 
-python:
-- "3.7"
-
 services:
+- docker
 - redis-server
 
 before_install:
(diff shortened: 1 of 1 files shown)


adarnimrod:
I see that you already merged this PR but I had a thought about the CI setup. If you already have Docker in the CI environment and the development setup is using containers, I'd be in favor of using the Redis and Postgres Docker images instead. This gives flexibility to use newer version that do not always have packages ready and it more closely follows the setup advocated.

bochecha:
> I'd be in favor of using the Redis and Postgres Docker images instead. This gives flexibility to use newer version that do not always have packages ready and it more closely follows the setup advocated.

Yes. In fact that's what I've been doing for development since day 1.

I'm a bit worried this makes the CI much longer though (having to pull the PostgreSQL/Redis Docker images each time, as they are bigger than just the few needed packages), pulling the Ubuntu one (the base for the Azafea image) already takes 2 minutes for each job. :sob: 

I wish there was a way to prepare the Travis host image in advance, so it already contains those Docker images pre-pulled…

adarnimrod:
Yeah, Travis does not allow caching Docker images (https://docs.travis-ci.com/user/caching/#things-not-to-cache). You can work around that with saving the images to the cache and loading them from the cache before pulling. There's an example here: https://github.com/travis-ci/travis-ci/issues/5358#issuecomment-248915326 .

PR #77, opened 2020-01-02
ci: Use containerized services in Travis

adarnimrod:
Try to improve the time it takes for a ci run by:
- Caching Docker images.
- Doing less for the lint job.

Don't merge this yet, the PR is for running the CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,6 @@
 dist: bionic
 sudo: required
 language: minimal
-addons:
-  apt:
-    sources:
-    - sourceline: "ppa:chris-lea/redis-server"
-    packages:
-    - redis-server
-    - redis-tools
 
 branches:
   only:
@@ -15,26 +8,16 @@ branches:
 
 services:
 - docker
-- redis-server
-
-before_install:
-- sudo apt --yes --quiet update
-
-# Use PostgreSQL 11
-- sudo apt --yes --quiet autoremove --purge libpq\* postgresql-{,client-,contrib-}{9,10}\*
-- sudo apt --yes --quiet install postgresql-11
 
 install:
 - docker build --build-arg build_type=dev --tag azafea-tests .
 
 before_script:
-# Create the PostgreSQL role and database
-- sudo -u postgres createuser azafea
-- sudo -u postgres createdb --owner=azafea azafea-tests
+- docker run -d --name=redis --network=host redis:5-alpine --requirepass 'CHANGE ME!!'
+- docker run -d --name=postgres --network=host -e 'POSTGRES_PASSWORD=CHANGE ME!!' -e 'POSTGRES_USER=azafea' -e 'POSTGRES_DB=azafea-tests' postgres:11-alpine
 
-# Set the default passwords for PostgreSQL and Redis, as that's what the tests expect
-- sudo -u postgres psql --command="ALTER USER azafea WITH PASSWORD 'CHANGE ME!!';"
-- redis-cli config set requirepass 'CHANGE ME!!'
+after_script:
+- docker rm -f redis postgres
 
 matrix:
   include:
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,18 +12,16 @@ services:
 install:
 - docker build --build-arg build_type=dev --tag azafea-tests .
 
-before_script:
-- docker run -d --name=redis --network=host redis:5-alpine --requirepass 'CHANGE ME!!'
-- docker run -d --name=postgres --network=host -e 'POSTGRES_PASSWORD=CHANGE ME!!' -e 'POSTGRES_USER=azafea' -e 'POSTGRES_DB=azafea-tests' postgres:11-alpine
-
-after_script:
-- docker rm -f redis postgres
-
 matrix:
   include:
   - name: "Integration Tests"
+    before_script:
+    - docker run -d --name=redis --network=host redis:5-alpine --requirepass 'CHANGE ME!!'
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,46 +1,43 @@
...
-- sudo -u postgres createdb --owner=azafea azafea-tests
+- docker build --build-arg build_type=dev --tag azafea-tests --pull .
 
-# Set the default passwords for PostgreSQL and Redis, as that's what the tests expect
-- sudo -u postgres psql --command="ALTER USER azafea WITH PASSWORD 'CHANGE ME!!';"
-- redis-cli config set requirepass 'CHANGE ME!!'
+before_cache:
+# Save Docker images to Travis' cache, but prune old ones first
+- mkdir -p docker_images
+- docker images --quiet --filter=dangling=true | xargs -r docker rmi
```
bochecha: Is this the same as `docker system prune`?

.travis.yml
```diff
@@ -1,46 +1,43 @@
...
+- docker save $(docker images -a -q) | gzip > docker_images/images.tar.gz
 
 matrix:
   include:
   - name: "Integration Tests"
+    before_script:
```
bochecha: I like that you're moving these here, because they are not necessary for the "lint" job and so that one can be faster. :+1:

Can I ask you to do that in a separate commit though, so it's easier to review and understand?

(i.e first move things here then do the caching stuff, or the contrary as you prefer)

.travis.yml
```diff
@@ -1,46 +1,43 @@
...
+    - docker image pull postgres:11-alpine
+    - docker image pull redis:5-alpine
```
bochecha: You use these ones because they are smaller? (that's fine, I just want to make sure I understand the "why")

.travis.yml
```diff
@@ -1,46 +1,43 @@
...
+- docker build --build-arg build_type=dev --tag azafea-tests --pull .
 
-# Set the default passwords for PostgreSQL and Redis, as that's what the tests expect
-- sudo -u postgres psql --command="ALTER USER azafea WITH PASSWORD 'CHANGE ME!!';"
-- redis-cli config set requirepass 'CHANGE ME!!'
+before_cache:
+# Save Docker images to Travis' cache, but prune old ones first
+- mkdir -p docker_images
+- docker images --quiet --filter=dangling=true | xargs -r docker rmi
+- docker save $(docker images -a -q) | gzip > docker_images/images.tar.gz
```
bochecha: I kind of wish you'd separate the speed up through caching from the move to Docker images.

(if you're not comfortable enough with Git to split and reorder commits, I'm happy to do it for you :slightly_smiling_face: )

.travis.yml
```diff
@@ -1,46 +1,43 @@
...
   - name: "Integration Tests"
+    before_script:
+    - docker image pull postgres:11-alpine
+    - docker image pull redis:5-alpine
+    - docker run -d --name=redis --network=host redis:5-alpine --requirepass 'CHANGE ME!!'
+    - docker run -d --name=postgres --network=host -e 'POSTGRES_PASSWORD=CHANGE ME!!' -e 'POSTGRES_USER=azafea' -e 'POSTGRES_DB=azafea-tests' postgres:11-alpine
     script:
     - docker run --rm --network=host --entrypoint="" azafea-tests pipenv run test-all
+    after_script:
+    - docker rm -f redis postgres
```
bochecha: Why remove the containers? Won't that make the images dangling, and as such they'll get removed in `before_cache` above?

.travis.yml
adarnimrod: Yeah, but this also works on older version of Docker. Doesn't matter anymore, I removed the whole cache thing (it made the build even slower).

.travis.yml
adarnimrod: Just for the smaller images (faster to download). There are actually tags for the latest Alpine variants. But this could be surprising (builds could suddenly fail if the image in Docker Hub is updated to a new major version).

.travis.yml
adarnimrod: You don't have to remove the containers (the VM is destroyed anyway). But the images from which the containers are made aren't touched.

adarnimrod:
I changed the PR from containers and speed up to just containers (and a little bit of a speed up). Only question I have is do you wish to the latest available version of Redis and Postgres or set versions?

.travis.yml
bochecha: Ah ok, thanks for the tip about older Docker then. :slightly_smiling_face:

.travis.yml
bochecha: > There are actually tags for the latest Alpine variants

Ah, they are caled "alpine"! I missed them because I was searching for something with "latest" in the name.

> But this could be surprising (builds could suddenly fail if the image in Docker Hub is updated to a new major version).

In general I agree with that.

For things like Redis and PostgreSQL though, I'm pretty confident about their consistent stability and compatibility that I'd be willing to risk it.

Although I guess it's nice to test with what is in production right now as well… :thinking: 

I know we use PostgreSQL 11 in production, are we also using Redis 5?

adarnimrod: We're on Redis 5.0. Since for now we're the only users of this we can set it to what Endless has in production. But for a more broad appeal we can actually run against multiple versions of PG and/or Redis (whatever we decide is supported). Personally I prefer to be explicit in this regard. I imagine that when we decide to update PG or Redis in production we'll start by adding a job to test on both versions (current and future) and keep that until all environments are updated to the new version and then we'll remove the old version from the CI.

.travis.yml
adarnimrod: Done.

.travis.yml
adarnimrod: Done.

.travis.yml
bochecha: > Since for now we're the only users of this we can set it to what Endless has in production.

Yes, let's.

PR #78, opened 2020-01-06
config: Allow passing PostgreSQL connection arguments

### LINKED ISSUE #74: Add database connection parameters configuration option
Allow setting database connection parameters from the configuration file (like `sslmode`).  We can get by for now with using environment variables that libpq uses but it would be nicer to use the configuration file. An option like `params` under the `postgresql` section that is a dictionary of the parameters is what I have in mind. Looking at https://docs.sqlalchemy.org/en/13/core/engines.html#custom-dbapi-args it seems like it's doable.

bochecha:
PostgreSQL's libpq supports many connection arguments:

https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-PARAMKEYWORDS

SQLAlchemy makes it easy to pass them down to PsycoPG2 and libpq:

https://docs.sqlalchemy.org/en/13/core/engines.html#custom-dbapi-args

This commit introduces a new "connect_args" configuration table, nested
under the "postgresql" one.

It can be passed arbitrary key/values, like "sslmode" or
"connect_timeout" for example.

Fixes #74

### DIFF
--- a/azafea/cli/commands.py
+++ b/azafea/cli/commands.py
@@ -61,8 +61,7 @@ def do_dropdb(config: Config, args: argparse.Namespace) -> None:
         log.error('Could not clear the database: no event queue configured')
         raise NoEventQueueExit()
 
-    db = Db(config.postgresql.host, config.postgresql.port, config.postgresql.user,
-            config.postgresql.password, config.postgresql.database)
+    db = Db(config.postgresql)
     db.drop_all()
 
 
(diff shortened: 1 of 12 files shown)


bochecha:
@adarnimrod with this, you'll be able to do the following in the configuration file:

```toml
[postgresql.connect_args]
sslmode = "require"
…
```

adarnimrod:
Thanks. That's exactly what I wanted.

PR #81, opened 2020-01-13
metrics: Deduplicate image version events

bochecha:
Looking at the database, we have *many* requests which contain *lots* of image version events.

The most I found in production is a request with 139 identical image version events.

This does two things:

1. it filters incoming requests so that only a single image version event (at most) is added per request;
2. it brings in a new CLI tool to deduplicate the existing image version records so that the DB only contains one per request;

### DIFF
--- a/azafea/event_processors/metrics/events/__init__.py
+++ b/azafea/event_processors/metrics/events/__init__.py
@@ -7,16 +7,18 @@
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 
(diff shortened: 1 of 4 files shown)


bochecha:
@adarnimrod once this is deployed, can you run the following Azafea command in production?

```
$ azafea metrics-2 dedupe-image-versions
```

Testing it here with data similar to prod, it took about 30 minutes to complete, and the peak required memory was a bit less than 1GB.

Issue #103, opened 2020-04-17
Docker image does not build


mcatanzaro:
Yo. With EOL of Ubuntu 19.04, the Docker image no longer builds:

```
$ podman build --tag azafea .
STEP 1: FROM ubuntu:disco
Getting image source signatures
Copying blob 0a4ccbb24215 done
Copying blob 4dc9c2fff018 done
Copying blob 5ff1eaecba77 done
Copying blob c0f243bc6706 done
Copying config c88ac1f841 done
Writing manifest to image destination
Storing signatures
STEP 2: ENV LANG C.UTF-8
03253aa050eac02e2a013a9d976efeca759419a3687e9415bd0763e8510fad59
STEP 3: WORKDIR /opt/azafea/src
32bcd4c990adefe5f48b33dc0e34fdc8ac2c0827e3752f101131906771e1a320
STEP 4: COPY Pipfile.lock .
4fa7b5c1b7438a8a0e627e2ed5f2a141d85cda003eb1c2022ffa0a11ef2afc0f
STEP 5: ARG build_type
63956f5e5f602095f3fbb575998a2ad1a17836569126ab1688a704fd3d6ceaad
STEP 6: RUN apt --quiet --assume-yes update &&     apt --quiet --assume-yes --no-install-recommends install         gcc         gir1.2-glib-2.0         gobject-introspection         libcairo2-dev         libffi-dev         libgirepository-1.0-1         libgirepository1.0-dev         libglib2.0-dev         libpq5         libpq-dev         python3         python3-dev         python3-pip         python3-setuptools         python3-wheel         &&     pip3 install pipenv &&     pipenv install --ignore-pipfile &&     if [ "${build_type}" = "dev" ]; then         pipenv install --ignore-pipfile --dev     ; else         apt --quiet --assume-yes autoremove --purge             gcc             libcairo2-dev             libffi-dev             libgirepository1.0-dev             libglib2.0-dev             libpq-dev             python3-dev             &&         rm -rf /var/cache/{apt,debconf}                /var/lib/apt/lists/*                /var/log/{apt,dpkg.log}                ~/.cache     ; fi
Ign:1 http://security.ubuntu.com/ubuntu disco-security InRelease
Err:2 http://security.ubuntu.com/ubuntu disco-security Release
  404  Not Found [IP: 2001:67c:1562::15 80]
Ign:3 http://archive.ubuntu.com/ubuntu disco InRelease
[...]

PR #106, opened 2020-04-21
Use the same Dockerfile as we use in production.

adarnimrod:
In production in Endless, we use a different Dockerfile with a slightly
different setup. The main points are using the Alpine variant of the
Python base image and rendering a `config.toml` file from a template and
environment variable. This commit uses (almost) exactly the same
Dockerfile that we use in production since Azafea was deployed and it
can be argued that it's the most tested one. We don't have a dev/ prod
image so that build argument is dropped.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ services:
 - docker
 
 install:
-- docker build --build-arg build_type=dev --tag azafea-tests .
+- docker build --tag azafea-tests .
 
 matrix:
   include:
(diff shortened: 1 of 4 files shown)


adarnimrod:
BTW, this will fix the breaking CI and #103.

wjt:
I do not have enough context to meaningfully review this

mcatanzaro:
I was able to avoid all problems by switching from podman to docker, and from Ubuntu Disco to Ubuntu Eoan.

Updating the Dockerfile for azafea-metrics-proxy would be good to do as well.

mcatanzaro:
> Updating the Dockerfile for azafea-metrics-proxy would be good to do as well.

endlessm/azafea-metrics-proxy/issues/12

PR #107, opened 2020-04-24
T28742 Further parental controls fixes

pwithnall:
Fix some incorrect parsing code from my last attempt, and add more tests.

https://phabricator.endlessm.com/T28742


pwithnall:
I should note that, as per https://phabricator.endlessm.com/T29910, this PR doesn’t actually fix all the cases where the `ParentalControlsChanged` event can crash due to parsing invalid types. It just fixes the one which we were hitting consistently.

Basically every unguarded call to `my_variant.get_{string,uint64,boolean,etc.}()` to get a value from an entry in an `a{sv}` needs to be fixed. I decided not to fix them for `ParentalControlsChanged` in this PR as that would involve all the groundwork for fixing T29910 (a not insignificant amount of work).

PR #109, opened 2020-04-30
Improve shell_app_is_open to allow easier and faster reports

liZe:
This PR brings two improvements:
- Add an index to improve speed when we want to retrieve data involving start
time and group it by app ID (7086bbf).
- Add a `duration` column allowing Metabase users to query duration with no SQL (0c2f39c).

`duration` is a float representing difference between `stopped_at` and `started_at` in seconds. It’s implemented as a SQLAlchemy default value, as the lines are not supposed to be changed.

https://phabricator.endlessm.com/T29220

### DIFF
--- a/azafea/event_processors/endless/metrics/events/__init__.py
+++ b/azafea/event_processors/endless/metrics/events/__init__.py
@@ -14,7 +14,7 @@ from gi.repository import GLib
 from sqlalchemy.dialects.postgresql import DOUBLE_PRECISION, JSONB
 from sqlalchemy.event import listens_for
 from sqlalchemy.inspection import inspect
-from sqlalchemy.schema import Column
+from sqlalchemy.schema import Column, Index
 from sqlalchemy.types import ARRAY, BigInteger, Boolean, Integer, LargeBinary, Numeric, Unicode
 
 from azafea.model import Base, DbSession
@@ -790,6 +790,11 @@ class ShellAppIsOpen(SequenceEvent):
     __payload_type__ = 's'
 
     app_id = Column(Unicode, nullable=False)
+
+    __table_args__ = (
+        Index('ix_shell_app_is_open_app_id_started_at', 'started_at', 'app_id',
+              postgresql_ops={'app_id': 'varchar_pattern_ops'}),
+    )
 
     @staticmethod
     def _get_fields_from_payload(payload: GLib.Variant) -> Dict[str, Any]:
--- a/azafea/event_processors/endless/metrics/events/__init__.py
+++ b/azafea/event_processors/endless/metrics/events/__init__.py
@@ -12,10 +12,20 @@ from typing import Any, Dict
 from gi.repository import GLib
 
 from sqlalchemy.dialects.postgresql import DOUBLE_PRECISION, JSONB
+from sqlalchemy.engine.default import DefaultExecutionContext
 from sqlalchemy.event import listens_for
 from sqlalchemy.inspection import inspect
 from sqlalchemy.schema import Column, Index
-from sqlalchemy.types import ARRAY, BigInteger, Boolean, Integer, LargeBinary, Numeric, Unicode
+from sqlalchemy.types import (
+    ARRAY,
+    BigInteger,
+    Boolean,
+    Float,
+    Integer,
+    LargeBinary,
+    Numeric,
+    Unicode,
+)
 
 from azafea.model import Base, DbSession
 from azafea.vendors import normalize_vendor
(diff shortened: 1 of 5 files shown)


azafea/event_processors/endless/metrics/events/__init__.py
```diff
@@ -784,12 +793,22 @@ def _get_fields_from_payload(payload: GLib.Variant) -> Dict[str, Any]:
...
 class ShellAppIsOpen(SequenceEvent):
     __tablename__ = 'shell_app_is_open'
     __event_uuid__ = 'b5e11a3d-13f8-4219-84fd-c9ba0bf3d1f0'
     __payload_type__ = 's'
 
     app_id = Column(Unicode, nullable=False)
+    duration = Column(Float, default=default_time_duration, nullable=False)
+
+    __table_args__ = (
+        Index('ix_shell_app_is_open_app_id_started_at', 'started_at', 'app_id'),
```
wjt, line 810: Is it possible to make the `app_id` field use `varchar_pattern_ops`? Based on https://www.postgresql.org/docs/9.4/indexes-opclass.html I believe this may be necessary to make queries like the following hit the index:

```sql
select * from shell_app_is_open
where started_at > now() - interval '30 days'
and app_id LIKE 'com.endlessm.encyclopedia.%'
```

though I would be equally happy to learn that the default operator class is enough to make postgres understand it can use the index for that prefix match.

liZe: > Is it possible to make the `app_id` field use `varchar_pattern_ops`?

You’re right, that would be useful.

> though I would be equally happy to learn that the default operator class is enough to make postgres understand it can use the index for that prefix match.

I’ve tried on dev data, and it looks like we have to explicitly set the option. Creating the index is slower but we get much faster requests. Hopefully SQLAlchemy provides a simple syntax for that.

liZe:
> I didn't do any testing, but the `shell_app_is_open` has close to 200M records and in this case we may want to move the population of the `duration` column to a separate task that we can perform while Azafea is running.

I’m currently importing 4 months of real data (it was already long to export…), I’ll have a rough estimation of the time needed for this small number of lines.

@adarnimrod If it’s too long (and it probably is), what’s the best way for you to perform this as a separate task? Should I put a default value in the deployment and let you launch SQL on the database to calculate the real values?

wjt:
Here's how it was done previously e3fd3df855d82c3629a1a90fd435d915bf25dddf

adarnimrod:
@liZe previously we added a new command to Azafea (like the commit @wjt mentioned). The workflow was doing the deployment like always (running migrations, replacing old containers with the new containers, then running the command in a separate container).

liZe:
> @liZe previously we added a new command to Azafea (like the commit @wjt mentioned). The workflow was doing the deployment like always (running migrations, replacing old containers with the new containers, then running the command in a separate container).

OK. I’m currently working on this new command and its tests.

Now that I have imported real data (it was incredibly long), I’m able to test it on a large number of lines.

azafea/event_processors/endless/metrics/events/__init__.py
liZe: I’ve added `varchar_pattern_ops`.

liZe:
I’ve added a separate command to set the duration time. I’m not really happy with the default value set to 0, but I can’t find a better solution (I don’t think that allowing NULL values is a good idea).

Chunks are used, as they are for other commands. I’m not sure that it’s as useful as it is for other commands, it may be really slower, I’ve not tested it on real data yet (but I can if you want, as soon as my `shell_app_is_open` is ready :unamused:).

@wjt @adarnimrod Let me know if I can improve this PR before merging!

adarnimrod:
@liZe looks good. The reason behind using chunks was to avoid running out of memory. We can set the default value to `null` and later on have another migration to disallow `null` values.

azafea/event_processors/endless/metrics/v2/migrations/d4809698244a_add_duration.py
```diff
@@ -0,0 +1,28 @@
...
+revision = 'd4809698244a'
+down_revision = '4a49897b47ea'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('shell_app_is_open', sa.Column('duration', sa.Float(), nullable=True))
+    op.execute('UPDATE shell_app_is_open SET duration = 0')
+    op.alter_column('shell_app_is_open', 'duration', nullable=False)
```
wjt, line 24: @adarnimrod does your last comment mean you think this should be split into different migrations?

adarnimrod: I meant that we can have `null` values in the interim before updating all of the records and afterwards change it so that `null` aren't accepted (this change will be in a separate PR). But I don't think that's necessary (0 or -1 is fine by me).

adarnimrod:
Deployed to the test and dev environments. Worked without an issue and the migrations worked correctly. Merging.

PR #111, opened 2020-05-11
Strip useless quotes and field from OS information

liZe:
OS information fields get useless extra quotes and field in the payload.

This pull request:
- ~~drops the useless `name` column,~~
- strips the useless quotes around `version` (and `name`) before inserting new rows,
- adds a command removing useless quotes from `version` (and `name`) in database.

~~This is a draft because:~~
- ~~it needs endlessm/eos-metrics-instrumentation#88 (draft too),~~
- ~~it includes an Alembic migration and thus has to be adjusted when #109 is merged.~~

https://phabricator.endlessm.com/T29322

### DIFF
--- a/azafea/event_processors/endless/metrics/events/__init__.py
+++ b/azafea/event_processors/endless/metrics/events/__init__.py
@@ -560,8 +560,8 @@ class OSVersion(SingularEvent):
     @staticmethod
     def _get_fields_from_payload(payload: GLib.Variant) -> Dict[str, Any]:
         return {
-            'name': payload.get_child_value(0).get_string(),
-            'version': payload.get_child_value(1).get_string(),
+            'name': payload.get_child_value(0).get_string().strip('"'),
+            'version': payload.get_child_value(1).get_string().strip('"'),
         }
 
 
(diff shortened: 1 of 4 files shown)


wjt:
> it needs endlessm/eos-metrics-instrumentation#88 (draft too),

I don't think it does. This code needs to cope with both the old and new submission formats, since there will be machines in the wild running old versions of the client code forever. As far as I can see, it does deal.

liZe:
> > it needs [endlessm/eos-metrics-instrumentation#88](https://github.com/endlessm/eos-metrics-instrumentation/pull/88) (draft too),
> 
> I don't think it does.

I wanted to say that we need to merge #88 sooner or later. We won’t forget now that it’s merged :smile:…

> I would ask to split the migration that drops the column to a separate PR that will deploy later. Everything else looks good to me.

I’ve put the migration into the `lize/drop-os-name` branch (including the commits from this PR). I’ve changed this PR to remove useless quotes from name too, even if the column is removed later.

adarnimrod:
Deployed to the test and dev environments. Worked without an issue and the migrations worked correctly. Merging.

PR #113, opened 2020-05-14
Store 2-letter country codes in database

liZe:
This PR includes these changes:
- DB migration allowing both alpha2 and alpha3 codes in PostgreSQL,
- SQLAlchemy changes allowing ~~only alpha2~~ alpha2 and alpha3 codes when inserting new records,
- CLI commands added for country code migration in `ping_v1` and `activation_v1`.

https://phabricator.endlessm.com/T29210

### DIFF
--- a/azafea/event_processors/endless/activation/v1/cli.py
+++ b/azafea/event_processors/endless/activation/v1/cli.py
@@ -10,12 +10,14 @@ import argparse
 import logging
 
 from sqlalchemy.orm.query import Query
+from sqlalchemy.sql.expression import func
 
 from azafea.config import Config
 from azafea.model import Db
 from azafea.utils import progress
 from azafea.vendors import normalize_vendor
 
+from ...country import transform_alpha_3_into_2
 from ...image import parse_endless_os_image
 from .handler import Activation
 
@@ -37,6 +39,14 @@ def register_commands(subs: argparse._SubParsersAction) -> None:
     parse_images.add_argument('--chunk-size', type=int, default=5000,
                               help='The size of the chunks to operate on')
     parse_images.set_defaults(subcommand=do_parse_images)
+
+    transform_countries_alpha_3_to_2 = subs.add_parser(
+        'transform-countries-alpha-3-to-2',
+        help='Transform 3-letter country codes to 2-letter country codes',
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    transform_countries_alpha_3_to_2.add_argument('--chunk-size', type=int, default=5000,
+                                                  help='The size of the chunks to operate on')
+    transform_countries_alpha_3_to_2.set_defaults(subcommand=do_transform_countries_alpha_3_to_2)
 
 
 def _normalize_chunk(chunk: Query) -> None:
@@ -98,5 +108,28 @@ def do_parse_images(config: Config, args: argparse.Namespace) -> None:
             progress(chunk_number * args.chunk_size, num_records)
 
     progress(num_records, num_records, end='\n')
+
+    log.info('All done!')
+
+
+def do_transform_countries_alpha_3_to_2(config: Config, args: argparse.Namespace) -> None:
+    db = Db(config.postgresql)
+    log.info('Transform alpha-3 country codes into alpha-2 for activations')
+
+    with db as dbsession:
+        query = dbsession.query(Activation).filter(func.length(Activation.country) == 3)
+        num_records = query.count()
+
+        if num_records == 0:
+            log.info('-> No activation with alpha-3 country code found')
+            return None
+
+        requests = transform_alpha_3_into_2(Activation.__tablename__)
+        for i, request in enumerate(requests, start=1):
+            dbsession.execute(request)
+            dbsession.commit()
+            progress(i, len(requests))
+
+    progress(len(requests), len(requests), end='\n')
 
     log.info('All done!')
--- a/azafea/event_processors/endless/country.py
+++ b/azafea/event_processors/endless/country.py
@@ -1,0 +1,53 @@
+# Copyright (c) 2020 - Endless OS LLC
+#
+# This file is part of Azafea
+#
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+from typing import List
+
+COUNTRIES = {
+    'AF': 'AFG', 'AL': 'ALB', 'DZ': 'DZA', 'AS': 'ASM', 'AD': 'AND', 'AO': 'AGO', 'AI': 'AIA',
+    'AQ': 'ATA', 'AG': 'ATG', 'AR': 'ARG', 'AM': 'ARM', 'AW': 'ABW', 'AU': 'AUS', 'AT': 'AUT',
+    'AZ': 'AZE', 'BS': 'BHS', 'BH': 'BHR', 'BD': 'BGD', 'BB': 'BRB', 'BY': 'BLR', 'BE': 'BEL',
+    'BZ': 'BLZ', 'BJ': 'BEN', 'BM': 'BMU', 'BT': 'BTN', 'BO': 'BOL', 'BQ': 'BES', 'BA': 'BIH',
+    'BW': 'BWA', 'BV': 'BVT', 'BR': 'BRA', 'IO': 'IOT', 'BN': 'BRN', 'BG': 'BGR', 'BF': 'BFA',
+    'BI': 'BDI', 'CV': 'CPV', 'KH': 'KHM', 'CM': 'CMR', 'CA': 'CAN', 'KY': 'CYM', 'CF': 'CAF',
+    'TD': 'TCD', 'CL': 'CHL', 'CN': 'CHN', 'CX': 'CXR', 'CC': 'CCK', 'CO': 'COL', 'KM': 'COM',
+    'CD': 'COD', 'CG': 'COG', 'CK': 'COK', 'CR': 'CRI', 'HR': 'HRV', 'CU': 'CUB', 'CW': 'CUW',
+    'CY': 'CYP', 'CZ': 'CZE', 'CI': 'CIV', 'DK': 'DNK', 'DJ': 'DJI', 'DM': 'DMA', 'DO': 'DOM',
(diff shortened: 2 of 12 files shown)


liZe:
This PR has been updated to allow both alpha2 and alpha3 codes for new rows. We can keep this version on production, update eos-activation-server to send alpha2 codes, and then merge the `lize-country-alpha-2-olny` branch.

liZe:
:warning: **We should wait for #112 to be merged before changing the Alembic version and merging this PR.** :warning:

liZe:
>  **We should wait for #112 to be merged before changing the Alembic deployment version and merging this PR.**

This PR includes migrations for ping and activation, and #112 includes a migration for metrics. It means that we don’t need to rebase these deployments, and that we can deploy #112 and #113 in any order.

azafea/event_processors/endless/country.py
```diff
@@ -0,0 +1,27 @@
...
+    'MEX': 'MX', 'MLT': 'MT', 'MNE': 'ME', 'MNP': 'MP', 'MOZ': 'MZ', 'MTQ': 'MQ', 'MYT': 'YT',
+    'NIU': 'NU', 'PAK': 'PK', 'PCN': 'PN', 'PLW': 'PW', 'PNG': 'PG', 'POL': 'PL', 'PRK': 'KP',
+    'PRT': 'PT', 'PRY': 'PY', 'PYF': 'PF', 'SEN': 'SN', 'SGS': 'GS', 'SLB': 'SB', 'SLV': 'SV',
+    'SPM': 'PM', 'SRB': 'RS', 'SUR': 'SR', 'SVK': 'SK', 'SVN': 'SI', 'SWE': 'SE', 'SWZ': 'SZ',
+    'SYC': 'SC', 'TCD': 'TD', 'TKM': 'TM', 'TUN': 'TN', 'TUR': 'TR', 'TUV': 'TV', 'UKR': 'UA',
+    'URY': 'UY', 'WLF': 'WF'}
+
+
+def tranform_alpha_3_into_2(alpha_3: str) -> str:
+    return COUNTRIES.get(alpha_3, alpha_3[:2])
```
wjt, line 27: The default here makes me very nervous. I think the transformation should fail for unknown country codes, rather than mapping them to some arbitrary pair of letters.

azafea/event_processors/endless/country.py
```diff
@@ -0,0 +1,27 @@
...
+    'CUW': 'CW', 'CYM': 'KY', 'DNK': 'DK', 'ESH': 'EH', 'EST': 'EE', 'FLK': 'FK', 'FRO': 'FO',
+    'FSM': 'FM', 'GIN': 'GN', 'GLP': 'GP', 'GNB': 'GW', 'GNQ': 'GQ', 'GRD': 'GD', 'GRL': 'GL',
+    'GUF': 'GF', 'GUY': 'GY', 'IRL': 'IE', 'IRQ': 'IQ', 'ISR': 'IL', 'JAM': 'JM', 'KAZ': 'KZ',
+    'KOR': 'KR', 'LBR': 'LR', 'LBY': 'LY', 'MAC': 'MO', 'MAF': 'MF', 'MDG': 'MG', 'MDV': 'MV',
+    'MEX': 'MX', 'MLT': 'MT', 'MNE': 'ME', 'MNP': 'MP', 'MOZ': 'MZ', 'MTQ': 'MQ', 'MYT': 'YT',
+    'NIU': 'NU', 'PAK': 'PK', 'PCN': 'PN', 'PLW': 'PW', 'PNG': 'PG', 'POL': 'PL', 'PRK': 'KP',
+    'PRT': 'PT', 'PRY': 'PY', 'PYF': 'PF', 'SEN': 'SN', 'SGS': 'GS', 'SLB': 'SB', 'SLV': 'SV',
+    'SPM': 'PM', 'SRB': 'RS', 'SUR': 'SR', 'SVK': 'SK', 'SVN': 'SI', 'SWE': 'SE', 'SWZ': 'SZ',
+    'SYC': 'SC', 'TCD': 'TD', 'TKM': 'TM', 'TUN': 'TN', 'TUR': 'TR', 'TUV': 'TV', 'UKR': 'UA',
+    'URY': 'UY', 'WLF': 'WF'}
```
wjt, line 23: It's rare but countries do come into existence and disappear again. Can we not use a library which exposes this data, rather than hardcoding a list ourselves? There is an `iso-codes` package, and https://pypi.org/project/pycountry/ seems a reasonably well-maintained wrapper around it.

azafea/event_processors/endless/country.py
```diff
@@ -0,0 +1,27 @@
+# Copyright (c) 2020 - Endless
```
wjt, line 1:
```suggestion
# Copyright (c) 2020 - Endless OS LLC
```

(and in all other new copyright headers)

azafea/event_processors/endless/country.py
liZe: It’s not arbitrary. The list of countries has been created from the ISO list, with only the list of 3-letter codes for countries whose 2-letter is different from the two first letters. All the values from the production database have been tested, and one was missing from the ISO list (a temporary code for Kosovo).

azafea/event_processors/endless/country.py
liZe: We could use an external package, but this code will be removed in the second step of the migration (when we’ll only have 2-letter codes everywhere), so there’s no need to take care of new countries appearing. Moreover, we’ll need a manual workaround for Kosovo that’s not included in the `iso-codes` package. I don’t think that relying on a Python package that needs a Debian package is really useful for some code that will be removed as soon as endlessm/eos-activation-server#23 is on production.

azafea/event_processors/endless/country.py
liZe: OK!

azafea/event_processors/endless/country.py
wjt: OK, I'd missed that this is a temporary mapping.

azafea/event_processors/endless/ping/v1/cli.py
```diff
@@ -122,3 +132,27 @@ def do_parse_images(config: Config, args: argparse.Namespace) -> None:
...
+
+        if num_records == 0:
+            log.info('-> No ping with alpha-3 country code found')
+            return None
+
+        for chunk_number, chunk in enumerate(query, start=1):
+            for ping in chunk:
+                ping.country = tranform_alpha_3_into_2(ping.country)
+            dbsession.commit()
+            progress(chunk_number * args.chunk_size, num_records)
```
wjt, line 154: Didn't we have an issue with another similar migration recently where this proved extremely slow? Might it perform better to do something like (pseudocode):

```
for alpha_3, alpha_2 in COUNTRIES.items():
    dbsession.execute("UPDATE activation_v1 SET country = ? WHERE country = ?", alpha_2, alpha_3)
```

which would allow the work to live in the database, rather than having to stream all 23 million rows (in chunks) in and out of the database?

wjt: I suppose we can try it and see.

wjt: Oh, I see @adarnimrod  had the same observation above. I would naively expect the `UPDATE` to perform better. I could be wrong!

liZe: > I suppose we can try it and see.

I can try on the test database.

> tranform_alpha_3_into_2

Oops, there’s a typo…

liZe: The migration takes about 3.5 days, it’s too long.

We can generate one request for each country value, that’s dirty but should be fast enough. I can try to find a better solution if you prefer.

wjt: I think one UPDATE per country code is a clean solution, personally :)

liZe: > I think one UPDATE per country code is a clean solution, personally :)

:smile:

The new solution takes 10 minutes for pings. Tests pass, the result seems to be OK, but I’d prefer to have another review, just to be sure…

azafea/event_processors/endless/country.py
```diff
@@ -0,0 +1,35 @@
...
+    'URY': 'UY', 'WLF': 'WF'}
+
+
+def transform_alpha_3_into_2(tablename: str) -> List[str]:
+    requests = [
+        f"UPDATE {tablename} SET country = '{alpha_2}' WHERE country = '{alpha_3}'"
+        for alpha_3, alpha_2 in COUNTRIES.items()]
+    requests.append(
+        f'UPDATE {tablename} SET country = substring(country FROM 1 FOR 2) '
+        'WHERE char_length(country) = 3')
```
wjt, line 34: I still don't really understand why this is preferable than making `COUNTRIES` a complete mapping from alpha-3 to alpha-2 for all known countries. (I believe that you have checked, of course!)

azafea/event_processors/endless/activation/v1/cli.py
```diff
@@ -100,3 +110,26 @@ def do_parse_images(config: Config, args: argparse.Namespace) -> None:
...
+        query = dbsession.query(Activation).filter(func.length(Activation.country) == 3)
+        num_records = query.count()
+
+        if num_records == 0:
+            log.info('-> No activation with alpha-3 country code found')
+            return None
+
+        requests = transform_alpha_3_into_2(Activation.__tablename__)
+        for i, request in enumerate(requests, start=1):
+            dbsession.execute(request)
```
wjt, line 129: Why use raw SQL here, instead of (something like, my sqlalchemy is rusty):

```python3
for alpha2, alpha3 in COUNTRIES:
    dbsession.execute(dbsession.query(Activation).update().where(Activation.country == alpha3).values(country=alpha2)
```

?

liZe: You’re right…

azafea/event_processors/endless/country.py
liZe: I can do this if you prefer.

wjt: It's a one-time migration so if this works, it's probably not worth spending much more time on it, but I would be more comfortable with this line removed and the COUNTRIES table updated to include all countries, yes.

azafea/event_processors/endless/activation/v1/cli.py
wjt: Again, a one-time migration so probably fine as-is, we can see by inspection that the queries as written are OK.

liZe: > I’ll check the prod data before resolving this conversation.

All the country codes from the ping and activation tables are in the list.

PR #115, opened 2020-06-04
Do not store empty values of location label information

liZe:
Empty values are sent by old versions of eos-metrics-instrumentation because gnome-initial-setup stores them in the configuration file by mistake. These empty strings are useless, we do not want to keep them in the database.

https://phabricator.endlessm.com/T29561

### DIFF
--- a/azafea/event_processors/endless/metrics/events/__init__.py
+++ b/azafea/event_processors/endless/metrics/events/__init__.py
@@ -37,6 +37,7 @@ from ..machine import (
 )
 from ..utils import get_asv_dict, get_bytes, get_child_values, get_strings
 from ._base import (  # noqa: F401
+    EmptyPayloadError,
     SequenceEvent,
     SingularEvent,
     # Reexport some symbols
@@ -458,7 +459,12 @@ class LocationLabel(SingularEvent):
 
     @staticmethod
     def _get_fields_from_payload(payload: GLib.Variant) -> Dict[str, Any]:
-        return {'info': get_asv_dict(payload)}
+        # Empty values are removed as they are useless, even if they are sent
+        # by old versions of eos-metrics-instrumentation
+        info = {key: value for (key, value) in get_asv_dict(payload).items() if value}
+        if not info:
+            raise EmptyPayloadError('Location label event received with no data.')
+        return {'info': info}
 
 
 class MissingCodec(SingularEvent):
--- a/azafea/event_processors/endless/metrics/events/_base.py
+++ b/azafea/event_processors/endless/metrics/events/_base.py
@@ -64,6 +64,7 @@ IGNORED_EVENTS: Set[str] = {
 IGNORED_EMPTY_PAYLOAD_ERRORS: Set[str] = {
     '9af2cc74-d6dd-423f-ac44-600a6eee2d96',
     'add052be-7b2a-4959-81a5-a7f45062ee98',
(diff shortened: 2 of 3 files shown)


azafea/event_processors/endless/metrics/events/__init__.py
```diff
@@ -458,7 +458,9 @@ class LocationLabel(SingularEvent):
 
     @staticmethod
     def _get_fields_from_payload(payload: GLib.Variant) -> Dict[str, Any]:
-        return {'info': get_asv_dict(payload)}
+        # Empty values are removed as they are useless, even if they are sent
+        # by old versions of eos-metrics-instrumentation
+        return {'info': {key: value for (key, value) in get_asv_dict(payload).items() if value}}
```
wjt, line 463: I don't think we want to store the row at all if it contains no useful information.

liZe: OK!

liZe: I can’t find an easy way to fully ignore the line if it’s empty.

The best solution I’ve found is to raise an `EmptyPayloadError` that will be catched by [`new_singular_event`](https://github.com/endlessm/azafea/blob/7fea174cefe4b29df4719bef149c49acc5a147c7/azafea/event_processors/endless/metrics/events/_base.py#L262). If the event is added to `IGNORED_EMPTY_PAYLOAD_ERRORS`, the line will silently be ignored (as far as I can understand).

Is there another better way?

wjt: That indeed seems to be the only currently-implemented way, and I don't see any obvious refactorings that would be clearer.

liZe: Done.

azafea/event_processors/endless/metrics/events/_base.py
```diff
@@ -64,6 +64,7 @@
 IGNORED_EMPTY_PAYLOAD_ERRORS: Set[str] = {
     '9af2cc74-d6dd-423f-ac44-600a6eee2d96',
     'add052be-7b2a-4959-81a5-a7f45062ee98',
+    'eb0302d8-62e7-274b-365f-cd4e59103983',
```
wjt, line 67:
```suggestion
    # LocationLabel
    'eb0302d8-62e7-274b-365f-cd4e59103983',
```

wjt: (I know the others don't have comments.)

Or could this fetch the UUID property from that class? I think not, because that would be a circular import.

liZe: > Or could this fetch the UUID property from that class? I think not, because that would be a circular import.

The circular import would be hard to avoid, I think that’s too much work for what we want to get.

But we can:
- Add a comment, even at the end of each line (event IDs have the same lengths, that would be readable).
- Fill `IGNORED_EVENTS` and `IGNORED_EMPTY_PAYLOAD_ERRORS` with the metaclass, like for the `*_EVENT_MODELS` dicts above.

I like the metaclass solution because it would put the information where it belongs: in the table classes. I could do that in another PR.

wjt: Yep, I agree that that could be a separate PR.

liZe: #117 can be merged after #115.

PR #116, opened 2020-06-11
Create materialized view listing machine ids grouped by day

liZe:
This view is useful to get active user reports.

We have to take a decision about the way the view is refreshed: an SQL command has to be sent regularly, probably once a day. @adarnimrod Do you have an idea?

https://phabricator.endlessm.com/T29216

### DIFF
--- a/azafea/cli/commands.py
+++ b/azafea/cli/commands.py
@@ -14,11 +14,13 @@ from alembic.command import revision as make_db_revision, upgrade as upgrade_db
 
 from redis import Redis
 from redis.exceptions import ConnectionError as RedisConnectionError
+from sqlalchemy.exc import ProgrammingError
 
 from ..config import Config
 from ..controller import Controller
 from ..migrations.utils import get_alembic_config, get_migration_heads, get_queue_migrations_path
-from ..model import Db, PostgresqlConnectionError
+from ..model import Db, PostgresqlConnectionError, views
+from ..utils import progress
 from .errors import ConnectionErrorExit, NoEventQueueExit, UnknownErrorExit
 
 
@@ -51,6 +53,10 @@ def register_commands(subs: argparse._SubParsersAction) -> None:
                                   'incoming one')
     replay.add_argument('queue', help='The name of the queue to replay, e.g "ping-1"')
     replay.set_defaults(subcommand=do_replay)
+
+    refresh = subs.add_parser('refresh-views',
+                              help='Refresh the content of the materialized views')
+    refresh.set_defaults(subcommand=do_refresh_views)
 
     run = subs.add_parser('run', help='Run azafea')
     run.set_defaults(subcommand=do_run)
@@ -181,6 +187,25 @@ def do_replay(config: Config, args: argparse.Namespace) -> None:
             raise UnknownErrorExit()
 
     log.info(f'Successfully moved failed events back to "{args.queue}"')
+
+
+def do_refresh_views(config: Config, args: argparse.Namespace) -> None:
+    db = Db(config.postgresql)
+    total_nb_views = nb_views = len(views)
+    log.info(f'Refreshing {nb_views} materialized views')
+
+    with db as dbsession:
+        for i, view in enumerate(views):
+            progress(i, nb_views)
+            try:
+                dbsession.connection().execute(f'REFRESH MATERIALIZED VIEW "{view}"')
+            except ProgrammingError as e:
+                assert 'UndefinedTable' in str(e)
+                log.debug(f'View {view} has not been created')
+                total_nb_views -= 1
+
+    progress(total_nb_views, total_nb_views, end='\n')
+    log.info(f'Successfully refreshed {total_nb_views} materialized views')
 
 
 def do_run(config: Config, args: argparse.Namespace) -> None:
--- a/azafea/event_processors/endless/metrics/request.py
+++ b/azafea/event_processors/endless/metrics/request.py
@@ -12,9 +12,9 @@ from hashlib import sha512
 from typing import Generator
 
 from sqlalchemy.schema import Column
-from sqlalchemy.types import BigInteger, DateTime, Integer, LargeBinary, Unicode
+from sqlalchemy.types import BigInteger, Date, DateTime, Integer, LargeBinary, Unicode
 
-from azafea.model import Base
+from azafea.model import Base, DbSession, View
 
 from gi.repository import GLib
 
@@ -32,6 +32,13 @@ class Request(Base):
     relative_timestamp = Column(BigInteger, nullable=False)
     machine_id = Column(Unicode(32), nullable=False)
     send_number = Column(Integer, nullable=False)
+
+
+class MachineIdsByDay(View):
+    __tablename__ = 'machine_ids_by_day'
+    __query__ = DbSession().query(
+        Request.received_at.cast(Date).label('day'),
+        Request.machine_id.label('machine_id')).distinct()
 
 
 class RequestBuilder:
--- a/azafea/event_processors/endless/metrics/tests/integration/test_metrics_cli.py
+++ b/azafea/event_processors/endless/metrics/tests/integration/test_metrics_cli.py
@@ -7,7 +7,7 @@
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 
 from gi.repository import GLib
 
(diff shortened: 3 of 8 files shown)


liZe:
If we want to keep the materialized view, I can also add a command to refresh them.

liZe:
Tests randomly fail, because the `initdb` command *sometimes* creates the materialized view after the parent table. I’ve added an explicit dependency relationship, but it’s not enough.

The table doesn’t appear in the metadata’s sorted tables list, because it’s not directly registered in the metadata. It’s registered in a temporary metadata object to avoid the real table creation. Listeners are then added to execute the `CREATE MATERIALIZED VIEW` statement.

The problem may also come from the way tables for different event processors are registered.

I have to find another solution, but I think that I’m not that far from a clean configuration for this view (and the next ones!)

(PS: [Original article about materialized views with SQLAlchemy](http://www.jeffwidman.com/blog/847/using-sqlalchemy-to-create-and-manage-postgresql-materialized-views/))

liZe:
It’s fixed. Take that, SQLAlchemy! :boom: 

> The problem may also come from the way tables for different event processors are registered.

The tables were cleared at the end of each tests, but not the events on which views are based.

I’ll also add a command refreshing materialized views.

liZe:
This pull request is ready for review.

The original goal of this pull request was to solve [T29216](https://phabricator.endlessm.com/T29216) by creating a fast way to get machine ids grouped by days.

A nice side effect is that we now have a nice background to create more materialized views:

- a new declarative SQLAlchemy class for materialized views,
- a new built-in command (`refresh-views`) to refresh materialized views,
- various new objects and functions to automatically create and delete materialized views for tests.

Even if the main goal is to provide SQL views for third-party tools, these views can also be used as any other tables by azafea.

New views could of course be useful to fix other slow requests, but they could also be a transitional solution for better anonymization. After we decide what to do in [T29292](https://phabricator.endlessm.com/T29292), we can create views to aggregate the data we don’t want to keep anymore. After a period of tests and reports from users, we can decide to replace views by real tables, keep the old table to store temporary data, and build requests that regularly aggregate and remove the old table data, and store them in the new one.

This PR is not perfect and could be improved (if needed) on various points, including:

- materialized views are not added to automatically generated Alembic deployments (like a lot of other SQLAlchemy corner cases),
- views are not stored in the metadata (where they should be), but in a module variable instead, because mypy doesn’t know the dynamically set `views` attribute,
- clearing metadata events uses a "private" `metadata.dispatch._clear()` method (current code already uses private methods and properties),
- the new command refreshing views silently fails when views are not created (just as `initdb` and `dropdb` do when tables are already created or already dropped), but should return an error when something wrong happens,
- tests cover the new code and check that everything works fine, but they’re not really elegant (in particular the ones in `test_cli.py`).

SQLAlchemy is not designed to work with views out-of-the-box, but it provides the needed APIs to handle them pretty well. We’ll have smaller and better code when they are fully supported.

azafea/event_processors/endless/metrics/v2/migrations/e7c7d5f2e4df_materialized_view_for_metrics.py
```diff
@@ -0,0 +1,27 @@
...
+# revision identifiers, used by Alembic.
+revision = 'e7c7d5f2e4df'
+down_revision = 'd4809698244a'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute(
+        "CREATE MATERIALIZED VIEW metrics_request_v2_by_day (month, machine_id) AS "
```
adarnimrod, line 22: Shouldn't `month` be called `date` or am I missing something?

azafea/model.py
```diff
@@ -247,6 +249,37 @@ class PostgresqlConnectionError(Exception):
 
 metadata = MetaData(naming_convention=NAMING_CONVENTION)
 Base = declarative_base(cls=BaseModel, constructor=BaseModel.__init__, metadata=metadata)
+views = {}
```
adarnimrod, line 252: I didn't see `views` being used anywhere.

azafea/event_processors/endless/metrics/v2/migrations/e7c7d5f2e4df_materialized_view_for_metrics.py
liZe: Of course…

azafea/model.py
liZe: It’s used in [`command.py`](https://github.com/endlessm/azafea/blob/e4b799c62479e39f7e0d31267d543b0dd3a987e3/azafea/cli/commands.py#L22).

azafea/event_processors/endless/metrics/request.py
```diff
@@ -34,6 +34,13 @@ class Request(Base):
     send_number = Column(Integer, nullable=False)
 
 
+class RequestByDay(View):
+    __tablename__ = 'metrics_request_v2_by_day'
```
wjt, line 38: May I bikeshed the name of this view? It's not a table of requests by day. It's a table of machine IDs by day. :)

azafea/event_processors/endless/metrics/request.py
```diff
@@ -34,6 +34,13 @@ class Request(Base):
...
+    __query__ = DbSession().query(
+        Request.received_at.cast(Date).distinct().label('day'),
```
wjt, line 40: My sqlalchemy is rusty but are we sure this evaluates to `SELECT DISTINCT received_at::date, machine_id` rather than `SELECT DISTINCT ON (received_at :: date) day, machine_id`, which would mean "for each distinct date, pick an arbitrary machine ID".

azafea/event_processors/endless/metrics/tests/integration/test_metrics_cli.py
```diff
@@ -1058,3 +1058,33 @@ def test_remove_os_info_no_quotes(self, capfd):
...
+
+        with self.db as dbsession:
+            assert dbsession.query(RequestByDay).count() == 0
+
+        self.run_subcommand('refresh-views')
+
+        with self.db as dbsession:
+            req1, req2 = dbsession.query(RequestByDay).order_by(RequestByDay.machine_id).all()
+            assert (req1.machine_id, req1.day) == ('machine1', occured_at.date())
+            assert (req2.machine_id, req2.day) == ('machine2', occured_at.date())
```
wjt, line 1090: I think this answers my question above about `DISTINCT` versus `DISTINCT ON`, but I think this test would be more meaningful if the request table contained rows with different dates (as opposed to three events with identical timestamps).

azafea/event_processors/endless/metrics/request.py
wjt: Perhaps `machine_ids_by_day`?

I could imagine a future iteration of this would want to handle the case where an offline machine comes online and submits 5 days' worth of events. We'd want the machine to be represented on all 5 days, not just the day it happens to be online. To do that we'd back the view by some other table (perhaps uptime). If the name is more generic, we won't need to rename it & update queries to do that.

azafea/cli/commands.py
```diff
@@ -52,6 +54,10 @@ def register_commands(subs: argparse._SubParsersAction) -> None:
     replay.add_argument('queue', help='The name of the queue to replay, e.g "ping-1"')
     replay.set_defaults(subcommand=do_replay)
 
+    refresh = subs.add_parser('refresh-views',
```
wjt, line 57: Did we ever work out where to trigger this? It will presumably need to run daily, soon after midnight UTC. (Are incoming requests processed in order? Do we know what the typical latency is? e.g. if it typically takes 5-10 mins for a request to hit the DB, the query should run no sooner than 00:10 UTC.)

adarnimrod: I'll add it to our deployment over on the `eos-administration` side. 00:30 UTC seems good enough to me.

azafea/event_processors/endless/metrics/request.py
liZe: > Perhaps `machine_ids_by_day`?

Much better.

> If the name is more generic, we won't need to rename it & update queries to do that.

I agree.

azafea/event_processors/endless/metrics/tests/integration/test_metrics_cli.py
liZe: > I think this answers my question above about `DISTINCT` versus `DISTINCT ON`, but I think this test would be more meaningful if the request table contained rows with different dates (as opposed to three events with identical timestamps).

Tests can definitely be improved, as said in my comment.

azafea/event_processors/endless/metrics/request.py
liZe: Done!

azafea/event_processors/endless/metrics/request.py
liZe: You’re right. The query is actually only used (for the moment) to get the tables that the view needs before it is created. The "real" request is in the deployment, and it’s not generated from this query (yet). It’s now fixed.

azafea/event_processors/endless/metrics/tests/integration/test_metrics_cli.py
liZe: I’ve improved this test, but feel free to add other comments if you think that it’s not good enough.

azafea/cli/commands.py
adarnimrod: BTW, the PR for the deployment side is at https://github.com/endlessm/eos-administration/pull/916.

adarnimrod:
@wjt are there any other concerns or can we proceed with this PR?

PR #118, opened 2020-07-02
Make tests pass regardless of PostgreSQL’s locale

liZe:


azafea/tests/integration/__init__.py
```diff
@@ -42,7 +42,8 @@ def ensure_no_tables(self):
                 with self.db as dbsession:
                     dbsession.query(model).all()
 
-            assert f'relation "{model.__tablename__}" does not exist' in str(exc_info.value)
+            assert model.__tablename__ in str(exc_info.value)
+            assert 'UndefinedTable' in str(exc_info.value)
```
wjt, line 46: I could be wrong but I think the exception will be an instance of this, which is a subclass of ProgrammingError – could we adjust the `with pytest.raises()` above to `with pytest.raises(UndefinedTable)`?

liZe: I’m not sure. I think that it’s a psycopg2 error embedded in an SQLAlchemy error, not a subclass, so `raises` won’t work.

liZe: It’s not an instance of `UndefinedTable`. I had already searched a long time ago for a way to get the original exception, I think that I didn’t find a clean solution. Did I miss something?

wjt: No, I'm probably just wrong :-)

wjt: (Unless it's a [chained exception](https://docs.python.org/3/reference/simple_stmts.html#raise) with `raise e1 from e2`?)

liZe: > (Unless it's a [chained exception](https://docs.python.org/3/reference/simple_stmts.html#raise) with `raise e1 from e2`?)

It’s not. But thank you for the link, I didn’t know chained exceptions!

dbnicholson:
I think this makes sense, but I'm still curious where the translated strings come from. If you run `LC_ALL=C.UTF-8 pytest` (or whatever the test command is) do you still have this issue? Because another way to approach this is to make sure the testing environment is consistent, and one thing to do there is to set a consistent locale. This can be done in a CI recipe, but it could also be done here using a wrapper script or something in pytest. https://subscription.packtpub.com/book/programming/9781789347562/4/ch04lvl1sec29/using-marks-from-fixtures is an example of adding a fixture to set the locale.

liZe:
> I think this makes sense, but I'm still curious where the translated strings come from.

They’re from the PostgreSQL server. The locale can be changed with the `lc_messages` value in `postgresql.conf`. If the value is not set, PostgreSQL uses the locale environment variable instead. I get the messages in `psql` too, they’re not related to Python libraries at all, and changing the locale for the tests doesn’t fix the problem.

PostgreSQL `initdb` command has an `--lc-messages=LOCALE` option, so I suppose that it’s possible to force the locale for a given database. A SQL query is also available for the administrator account, but it only works for the current session as far as I can tell.

My value has always been `fr_FR` as far as I can remember, no idea why the tests suddenly started to fail. Setting `lc_messages` in the PostgreSQL container (if it’s not already done) is probably the easiest way to have consistent results.

dbnicholson:
Oh interesting. I didn't think the server would send messages at all. I'd think it would send error codes and the client would provide strings for those. That does seem more difficult to handle since the database setup is not part of the test fixture.

PR #121, opened 2020-07-14
Add new event tables

### LINKED ISSUE #64: Figure out storing very big integers in PostgreSQL
Some of the Endless metrics events have values as unsigned 64 bits integers.

The biggest integer PostgreSQL can store, `bigint` is a signed 64 bits integer, which is too small to store those.

We have a few possibilities, which need to be investigated, off the top of my head:

1.  pretend they are signed 64 bits integers: we could just store them anyway as is; the value in DB would be wrong, but it could be cast when stored/retrieved; that would make queries harder?

2.  store them as binary blobs (`bytea`): we could just store the binary representation of the numbers (e.g 10 would be stored as `1010`); that would make queries harder?

3.  store them as strings: we could store the string representation of the numbers (e.g 10 would be stored as `"10"`); queries wouldn't be too hard, it would just be frustrating to always remember to use quotes;

4.  store them as `numeric(20, 0)`: this type allows number of arbitratry size (up to 1000 digits!) with exact precision (as opposed to floats); however calculations with such numbers are very slow compared to integers;

There might be other possibilities…

liZe:
(WIP: tests missing.)

Fix #64.


azafea/event_processors/endless/metrics/events/__init__.py
```diff
@@ -791,6 +791,48 @@ def _get_fields_from_payload(payload: GLib.Variant) -> Dict[str, Any]:
...
+    __payload_type__ = '(tt)'
+
+    # These come in as uint64, but the values won’t reach the limit of a BIGINT (int64)
+    number_of_valid_elements = Column(BigInteger, nullable=False)
+    number_of_bytes_read = Column(BigInteger, nullable=False)
+
+    @staticmethod
+    def _get_fields_from_payload(payload: GLib.Variant) -> Dict[str, Any]:
+        return {
+            'number_of_valid_elements': payload.get_child_value(0).get_uint64(),
```
wjt, line 806: What would happen if an event was submitted with a value greater than the upper bound of an int64? Presumably undesirable/unpredictable results.

Rather than just assuming it won't happen, let's define a function like:

```python3
def clamp_to_int64(u64):
    return min(u64, GLib.MAXINT64)
```

and then:

```suggestion
            'number_of_valid_elements': clamp_to_int64(payload.get_child_value(0).get_uint64()),
```

etc.

azafea/event_processors/endless/metrics/events/__init__.py
```diff
@@ -791,6 +791,48 @@ def _get_fields_from_payload(payload: GLib.Variant) -> Dict[str, Any]:
         return {'tables': payload.get_uint32()}
 
 
+class CacheHasInvalidElements(SingularEvent):
+    __tablename__ = 'cache_has_invalid_elements'
+    __event_uuid__ = 'cbfbcbdb-6af2-f1db-9e11-6cc25846e296'
+    __payload_type__ = '(tt)'
+
+    # These come in as uint64, but the values won’t reach the limit of a BIGINT (int64)
```
wjt, line 799: We should have a justification in the comment here (and below).

azafea/event_processors/endless/metrics/events/__init__.py
liZe: > Rather than just assuming it won't happen, let's define a function like:

Of course! It’s just a draft PR for now, I don’t even know if it works with real data!

azafea/event_processors/endless/metrics/events/__init__.py
liZe: :+1:

liZe:
:heavy_check_mark: replaying unknown data works!

I have to know why the size of read data is a huge value. Looks more like a token (it’s [advertised as a token in eos-event-recorder-daemon](https://github.com/endlessm/eos-event-recorder-daemon/blob/73df4a08387a678792a1d117d5560d749b59d5ca/daemon/emer-circular-file.c#L722-L733)) than a data size…

liZe:
> I have to know why the size of read data is a huge value. Looks more like a token (it’s [advertised as a token in eos-event-recorder-daemon](https://github.com/endlessm/eos-event-recorder-daemon/blob/73df4a08387a678792a1d117d5560d749b59d5ca/daemon/emer-circular-file.c#L722-L733)) than a data size…

@adarnimrod @dbnicholson I think that I’ve finally found where the problem comes from.

https://github.com/endlessm/eos-event-recorder-daemon/blob/bf6b292b2088d41bd3445c95d4f34f9a38d444b2/daemon/emer-daemon.c#L785

Shouldn’t we use `*token` instead of `token`?

liZe:
This PR is ready for review.

Tests have been added, they check the clamping function too.

Replaying old data works and shouldn’t take too long on production with enough workers. The code automatically replays only known events, there’s no need to manually filter them.

From the sample of old data I’ve tested, the content of the columns is OK. The only problem we have is the size of data read from cache, but that’s a bug in eos-event-recorder-daemon as far as I can tell. It means that all the data we have so far is meaningless for this value, and there’s no way to get it back. I can fix the bug in eos-event-recorder-daemon, if you think that it’s a bug too.

You can join me today, I’ll read my messages on Slack next week too.

wjt:
@liZe great catch. I opened a PR to fix the `*token` typo and also another type error by inspection (but I haven't even compiled it!). https://github.com/endlessm/eos-event-recorder-daemon/pull/239

I think we just accept that many old events have garbage in that field. It would be interesting to know whether any events have garbage in the first field, given the other bug I think I found :)

wjt:
@adarnimrod could you take a look at the details here?

liZe:
> It would be interesting to know whether any events have garbage in the first field, given the other bug I think I found :)

After 35,000 events replayed (100 lines added to `cache_has_invalid_elements`), all the values are good. I let the script run, I’ll give more reliable information when I have more lines.

liZe:
> I let the script run, I’ll give more reliable information when I have more lines.

Replaying old events is really, really slow.

After 100,000 events (200 lines in `cache_has_invalid_elements`), there’s no obviously wrong value. I’ll modify the script to only replay cache events and run it one more night.

azafea/event_processors/endless/metrics/utils.py
```diff
@@ -142,3 +142,7 @@ def get_event_datetime(request_absolute_timestamp: int, request_relative_timesta
     event_absolute_timestamp_sec = event_absolute_timestamp / 1000000000
 
     return datetime.fromtimestamp(event_absolute_timestamp_sec, tz=timezone.utc)
+
+
+def clamp_to_int64(u64: int) -> int:
```
adarnimrod, line 147: IIRC, currently there aren't cases where the values are needed/ more than 64-bit. However, I feel that some logging on those cases would be nice. I'm not sure what or how, but I've got a feeling it can help debugging further down the line. I don't see this issue as a blocker so I'm happy with creating a task to handle this later and proceed with this PR.

adarnimrod (changes requested):
LGTM. I had a single request that is not a blocker.

azafea/event_processors/endless/metrics/utils.py
liZe: > I feel that some logging on those cases would be nice. I'm not sure what or how, but I've got a feeling it can help debugging further down the line.

I agree, I can try to find a simple way to log this.

liZe:
> I’ll modify the script to only replay cache events and run it one more night.

After 1 night I’ve inserted almost 30,000 lines into `cache_has_invalid_elements`. 170 lines have a bad value for `number_of_valid_elements`, you were right @wjt!

I don’t know what to do with these wrong values. We can probably fix them later.

The replay also crashed:

```
Traceback (most recent call last):
  File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/lize/Informatique/azafea/azafea/__main__.py", line 16, in <module>
    cli.run_command(*sys.argv[1:])
  File "/home/lize/Informatique/azafea/azafea/cli/__init__.py", line 63, in run_command
    args.subcommand(config, args)
  File "/home/lize/Informatique/azafea/azafea/event_processors/endless/metrics/v2/cli.py", line 450, in do_replay_unknown
    dbsession.commit()
  File "/home/lize/.local/share/virtualenvs/azafea-_3hjtXBi/lib/python3.8/site-packages/sqlalchemy/orm/session.py", line 1036, in commit
    self.transaction.commit()
  File "/home/lize/.local/share/virtualenvs/azafea-_3hjtXBi/lib/python3.8/site-packages/sqlalchemy/orm/session.py", line 503, in commit
    self._prepare_impl()
  File "/home/lize/.local/share/virtualenvs/azafea-_3hjtXBi/lib/python3.8/site-packages/sqlalchemy/orm/session.py", line 471, in _prepare_impl
    self.session.dispatch.before_commit(self.session)
  File "/home/lize/.local/share/virtualenvs/azafea-_3hjtXBi/lib/python3.8/site-packages/sqlalchemy/event/attr.py", line 261, in __call__
    fn(*args, **kw)
  File "/home/lize/Informatique/azafea/azafea/event_processors/endless/metrics/events/__init__.py", line 893, in receive_before_commit
    upsert_machine_demo(dbsession, instance.request.machine_id)
AttributeError: 'NoneType' object has no attribute 'machine_id'
```

I had to add a dirty workaround to avoid the crash, but there’s probably a much better fix to find. @adarnimrod Did you see this traceback before?

liZe:
> ```
>     upsert_machine_demo(dbsession, instance.request.machine_id)
> AttributeError: 'NoneType' object has no attribute 'machine_id'
> ```

I don’t understand how it’s possible to have `None` in `instance.request`. Unknown lines all have a `request_id` that is for sure in the `metrics_request_v2` table, as there’s a foreign key.

azafea/event_processors/endless/metrics/utils.py
liZe: I’ve added an error, as it’s done elsewhere in azafea. I’m not sure that it’s really useful, but it will be at least displayed when events are replayed, and that’s where most of the problems should be caught.

adarnimrod:
> @adarnimrod Did you see this traceback before?

No, didn't see an error like that before.

adarnimrod (approved):

liZe:
I’ve tried to reproduce the error, but I’ve the script for multiple nights without a crash. There’s nothing strange in the database. It was probably a network or a database problem.

I think that we can launch the script and keep an eye on it. Even if it crashes, the table won’t be in a bad shape. It would be annoying for sure, but not that bad.

liZe:
Thanks to the new backup, I’ve been able to reproduce the bug and to debug. It will for sure happen on production if it’s not fixed.

We have a crash because the relationship attribute `request` is not set on this new object. When ORM objects are created, their relationships are generally lazy-loaded when needed. But in this case, `request` is not set and gives `None` when accessed. 

Flushing the session sets the `request` attribute and fixes the bug. But that’s not the best solution.

SQLAlchemy enables lazy loading for relationships when possible, but it’s obviously not possible on this case. I have to find why, but it’s complicated: the `request` relationship is not a classical SQL relationship, because it’s set on all the tables with a foreign key on the `Request` table. To avoid the declaration of the attribute on each table, it’s set on the `MetricEvent` abstract class with a `declared_attr` decorator. The `request_id` foreign key is a `declared_attr` too, but it works in our case because it’s set during the ORM object creation…

I’ll try different changes to understand why the relationship is not available, and to find out why it only happens with this peculiar row.

liZe:
I think that the problem is now solved.

Relationship attributes are not available until the session is flushed or committed. We have to use [`enable_relationship_loading`](https://docs.sqlalchemy.org/en/13/orm/session_api.html#sqlalchemy.orm.session.Session.enable_relationship_loading) to lazily load the relationships even if the row has not been inserted yet. The commit message includes more information about this.

I’m not sure that repeating the `enable_relationship_loading` call is the best solution, but it’s at least readable and doesn’t enable the feature when we don’t require it.

adarnimrod:
Do we have a Phabricator task for this?

wjt:
Created https://phabricator.endlessm.com/T30987

PR #122, opened 2020-09-07
Add command removing events with empty location info

liZe:
Events with empty location info are now ignored, but they used to be accepted and put in the database. This command removes these useless empty locations from the database.

Fix https://phabricator.endlessm.com/T29561. Related to #115.

### DIFF
--- a/azafea/event_processors/endless/metrics/v2/cli.py
+++ b/azafea/event_processors/endless/metrics/v2/cli.py
@@ -23,6 +23,7 @@ from ..events import (
     DualBootBooted,
     ImageVersion,
     LiveUsbBooted,
+    LocationLabel,
     InvalidAggregateEvent,
     InvalidSequence,
     InvalidSingularEvent,
@@ -130,6 +131,14 @@ def register_commands(subs: argparse._SubParsersAction) -> None:
         help='Remove leading and trailing quotes from OS information',
         formatter_class=argparse.ArgumentDefaultsHelpFormatter)
     remove_os_info_quotes.set_defaults(subcommand=do_remove_os_info_quotes)
+
+    remove_empty_location_info = subs.add_parser(
+        'remove-empty-location-info',
+        help='Remove location events with empty info',
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    remove_empty_location_info.add_argument('--chunk-size', type=int, default=5000,
+                                            help='The size of the chunks to operate on')
+    remove_empty_location_info.set_defaults(subcommand=do_remove_empty_location_info)
 
 
 def do_dedupe_dualboots(config: Config, args: argparse.Namespace) -> None:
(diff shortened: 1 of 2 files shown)


liZe:
This PR could be improved, but I don’t think that it would be really useful.

1. We compare `info` to `{'id': '','city': '','state': '','street': '','country': '','facility': ''}`. We could filter on dictionaries where values are empty-ish, but that would require a more complex request. We only have `{'id': '','city': '','state': '','street': '','country': '','facility': ''}` in database according to the Phabricator issue and to what I have in the backup database, I think that it’s useless to have a better request.

2. We use chunks, but it’s slow. I can change this if needed, as we have to remove more than 35 million lines.

3. Tests are simple, they cover all the code lines and the common use cases, but I can improve them.

azafea/event_processors/endless/metrics/v2/cli.py
```diff
@@ -532,3 +541,36 @@ def do_remove_os_info_quotes(config: Config, args: argparse.Namespace) -> None:
...
+
+        if num_records == 0:
+            log.info('-> No locations events with empty info in database')
+            return None
+
+        progress(0, num_records)
+
+        for chunk_number, chunk in enumerate(query, start=1):
+            for record in chunk:
+                dbsession.delete(record)
```
wjt, line 570: Won't this fetch every one of the records from the DB in order to delete it? Wouldn't

```
DELETE FROM ... WHERE ...
```

be a better approach than issuing `n / chunk_size` `SELECT`s and `n` `DELETE`s?

wjt: I agree with the general point that it is not worth spending too much time improving this job which will be run once. I am just concerned that 7 000 `SELECT`s and 35 000 000 `DELETE`s will take long enough that it would be an improvement to write the SQLAlchemy equivalent of:

```sql
DELETE FROM location_label
WHERE info @@ '{"id": "", "city": "", "state": "", "street": "", "country": "", "facility": ""}'
```

and also less code to write? (I would hope this would be better but I am not an expert!)

liZe: > I am just concerned that 7 000 `SELECT`s and 35 000 000 `DELETE`s will take long enough that it would be an improvement to write the SQLAlchemy equivalent of:

I agree, that would be both easier and faster, that’s my point 2. I usually use chunks (as it’s done everywhere else) and discuss with @adarnimrod, but it’s probably obvious here.

liZe:
> * We use chunks, but it’s slow. I can change this if needed, as we have to remove more than 35 million lines.

Edit: it’s more than **3** million.

azafea/event_processors/endless/metrics/v2/cli.py
adarnimrod: I would prefer the plain SQL over the SQLAlchemy version for the improved runtime as well. The progress bar is good to have when we have commands that take a considerable amount of time.

liZe: > I would prefer the plain SQL over the SQLAlchemy version for the improved runtime as well. The progress bar is good to have when we have commands that take a considerable amount of time.

OK. I think that I can write an SQLAlchemy request that will generate pure SQL and thus have the same runtime, but I’ll use plain SQL if the solution is not obvious.

liZe: The whole script removing the 3 million lines took 20 seconds, the SQLAlchemy request was already pure SQL launched on the server. Removing the chunks was enough.

adarnimrod (approved):
LGTM.

PR #124, opened 2020-09-16
metrics: Process "underscan enabled" events

mwleeds:
This commit is "inspired" by the following one:
metrics: Process the "demo mode entered" events

https://phabricator.endlessm.com/T30766


liZe (approved):
It’s OK for me.

PR #127, opened 2020-10-22
Migrate event documentation from wiki to Azafea

liZe:
https://phabricator.endlessm.com/T30730

All the documentation from the [wiki page](https://phabricator.endlessm.com/w/software/metrics_events/currently_recorded_events/) has been migrated to Azafea’s documentation, including:

- events currently stored (in docstrings),
- events unknown for Azafea (in Sphinx),
- deprecated events (in Sphinx).

The result is already available [on ReadTheDocs](https://azafea.readthedocs.io/en/lize-documentation/events.html) (temporary URL as it’s for a branch and not master, the final URL will be https://azafea.readthedocs.io/).

When this PR is merged, the documentation can automatically be pushed on ReadTheDocs with Travis, and the wiki page removed.


adarnimrod (approved):
LGTM.

docs/source/events.rst
```diff
@@ -0,0 +1,448 @@
+Events
+======
+
+This page is a list of events that Endless currently records for users that
+have not opted out of metrics recording.
```
wjt, line 5:
```suggestion
This page is a list of events that Endless OS currently records for users that
have not opted out of metrics recording, as well as events recorded by previous
versions of Endless OS which have been removed.
```

perhaps?

wjt (approved):
One superficial suggestion!

Reading through some of these, I think we could do with moving even more documentation into these event classes. For instance, a few of the events are only sparsely documented here, and the full semantics are defined in the code where they're recorded. We can do this incrementally once this change lands.

docs/source/events.rst
wjt: @liZe ^?

wjt (changes requested):

docs/source/events.rst
liZe: Fixed.

wjt (approved):

.travis.yml
```diff
@@ -25,3 +25,6 @@ matrix:
   - name: "Lint and Type Checking"
     script:
     - docker run --rm --network=host --entrypoint="" azafea-tests pipenv run lint
+  - name: Update ReadTheDocs
+    script:
+    - docker run --rm azafea-tests deploy-documentation --token "$RTD_TOKEN"
```
wjt, line 30: Is it correct to run this on pull requests, as is currently done? e.g. https://travis-ci.org/github/endlessm/azafea/jobs/746713046

adarnimrod: I think we can merge this one. There isn't anything to deploy (except for the token that's already in place).

wjt: What happens when this trigger runs on a pull request? Does the PR's docs get deployed to RTD?

adarnimrod: I now understand what you meant in the previous comment. Yes, it would run but it would re-deploy the `master` branch.

wjt:
@adarnimrod @liZe is there any reason to not hit merge on this? :)

wjt:
Based on @adarnimrod's comment above I crossed my fingers and hit merge.

PR #129, opened 2020-11-09
Add a sphinx module to upload documentation on metabase

liZe:
See https://phabricator.endlessm.com/T30895.

**This PR has to be merged after #127.**

The documentation for known tables and their fields can be uploaded using:
~`sphinx-build docs/source/ docs/build/html/ -b metabase -D metabase_session=<key> -D metabase_url=<url> -E`~
`sphinx-build docs/source/ docs/build/html/ -b metabase -D metabase_username=<username> -D metabase_password=<password> -D metabase_url=<url> -E`

Environment variables (uppercase) can also be used instead of Sphinx properties.

This code uses Sphinx to gather docstrings from the documentation, translate them from ReST to plain text, and upload them on Metabase. It suffers from many problems that can be solved in other pull requests later if needed.

- Transforming ReST to plain text is painful. Docutils doesn’t include a plain text generator, we have to use one provided by Sphinx. Unfortunately, this generator is not a "real" Docutils generator: it’s only designed to work with Sphinx, not with Docutils, and it crashes when it meets Sphinx features in ReST. That’s why we have to manually remove most of the Sphinx-specific notations. Using other solutions (Pandoc?) may give better results with less code.

- We don’t have documentation for ping events, activation events, unknown events and unknown tables.

- We don’t have documentation for many fields, including IDs, fields shared by many tables, foreign keys…

- Getting information from Metabase and uploading documentation is slow (~15 minutes for me on the dev platform through VPN). We have to launch one HTTP request per field, because information provided elsewhere (by routes that give information for many fields) are not enough. We could improve this by launching parallel HTTP requests. Or, maybe we don’t care :smile:.

- The only way I found to get a metabase authetication token is to login with my account and get the session token from the HTTP headers. There’s probably a better way to do this.

You can see the result on the dev platform: https://metabase.dev.endlessm-sf.com/browse/2

### DIFF
--- a/docs/source/metabase.py
+++ b/docs/source/metabase.py
@@ -1,0 +1,188 @@
+"""
+A Sphinx builder putting documentation on Metabase.
+
+"""
+
+import os
+import re
+from typing import Any, Dict, List
+
+import sphinx.writers.text
+import requests
+from docutils.core import publish_string
+from sphinx.application import Sphinx
+from sphinx.builders.dummy import DummyBuilder
+from sphinx.util import logging
+from sphinx.writers.text import TextTranslator, TextWriter
+
+logger = logging.getLogger(__name__)
+
+# Such a shame to depend on globals…
+sphinx.writers.text.MAXWIDTH = float('inf')
+
+
+class MetabaseBuilder(DummyBuilder):
+    """Sphinx builder used to deploy descriptions on Metabase."""
+    name = 'metabase'
+    default_translator_class = TextTranslator
+    tables = None
+    writer = None
+    _session = None
+
+    def _get(self, path: str) -> Dict[str, Any]:
+        """Launch a GET requests on Metabase API."""
+        with self.session.get(f'{self.config.metabase_url}/{path}') as response:
+            response.raise_for_status()
+            return response.json()
+
+    def _put(self, path: str, json: Dict[str, Any]) -> Dict[str, Any]:
+        """Launch a PUT requests on Metabase API."""
+        with self.session.put(f'{self.config.metabase_url}/{path}', json=json) as response:
+            response.raise_for_status()
+            return response.json()
+
+    @property
+    def session(self):
+        if not self._session:
+            self._session = requests.Session()
+
+            # Gather authentication headers
+            json = {
+                'username': self.config.metabase_username,
+                'password': self.config.metabase_password,
+            }
+            with self._session.post(f'{self.config.metabase_url}/session', json=json) as response:
+                response.raise_for_status()
+                self._session.headers.update({
+                    'Content-Type': 'application/json',
+                    'X-Metabase-Session': response.json()['id'],
+                })
+
+        return self._session
+
+    def format_description(self, lines: List[str]) -> str:
+        """Format tables and fields descriptions from ReST to plain text."""
+
+        # Unwrap lines
+        unwrapped_lines = []
+        for line in lines:
+            empty = not unwrapped_lines or not unwrapped_lines[-1] or not line
+            list_item = line and line.startswith('- ')
+            block_start = (
+                line and line.startswith('  ') and
+                not empty and not unwrapped_lines[-1].startswith('  '))
+            if empty or list_item or block_start:
+                unwrapped_lines.append(line)
+            else:
+                unwrapped_lines[-1] += ' ' + line.lstrip()
+
+        # Manually format and remove useless information
+        clean_lines = []
+        for line in unwrapped_lines:
+            line = line.replace(':class:', '')
+            if not line:
+                if clean_lines and clean_lines[-1]:
+                    clean_lines.append(line)
+                continue
+            elif re.match('^:[A-Za-z ]*:', line):
+                continue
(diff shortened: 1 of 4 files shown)


dbnicholson:
> * The only way I found to get a metabase authetication token is to login with my account and get the session token from the HTTP headers. There’s probably a better way to do this.

I didn't review any of this, but that statement seemed odd. Indeed, it seems the only way to use the REST API is via a session token - https://github.com/metabase/metabase/wiki/Using-the-REST-API. That's kinda odd for these types of services, but not insurmountable. What I suggest is we create a dedicated user with standard username/password credentials and then make those credentials available to this code so that it can generate a session token as needed. Whether this is done via travis or our jenkins, I'd allow those credentials to come from environment variables as that's how most CI systems would expose them to you.

docs/source/metabase.py
```diff
@@ -0,0 +1,174 @@
...
+            f'{self.config.metabase_url}/{path}', headers=self.headers, json=json)
+        response.raise_for_status()
+        return response.json()
+
+    @property
+    def headers(self) -> Dict[str, Any]:
+        """Get headers used for requests on Metabase API."""
+        return {
+            'Content-Type': 'application/json',
+            'X-Metabase-Session': self.config.metabase_session,
```
dbnicholson, line 48: Since the session token expires, I think you'll need to add some code to get one from `/api/session` using normal credentials. I also think you'll want to have a way to get the credentials from environment variables. It kinda seems like this is coming from a sphinx config file, but I'm not familiar with that interface.

liZe: Of course, that would be much easier. We’ll store the login and the password as environment variables in the CI, and give them to the CLI. They’re hidden in the logs, at least with Travis.

dbnicholson: I would still suggest getting them out of environment variables in azafea instead of marshaling them in as CLI arguments. It's less secure and adds more failure cases to handle. It would probably ultimately be less code than marshaling them through the CLI. Not a showstopper, though.

liZe: Authentication now uses username/password instead of session token. Properties can be given as environment variables or Sphinx properties (useful when developing, doesn’t require much more code).

docs/source/metabase.py
```diff
@@ -0,0 +1,186 @@
...
+    default_translator_class = TextTranslator
+    tables = None
+    writer = None
+    session = None
+
+    def _get(self, path: str) -> Dict[str, Any]:
+        """Launch a GET requests on Metabase API."""
+        response = requests.get(f'{self.config.metabase_url}/{path}', headers=self.headers)
+        response.raise_for_status()
+        return response.json()
```
dbnicholson, line 36: Although the connection will eventually be closed with python deletes the response object, I think it's cleaner to use the response as a context manager so it's closed immediately. Like:
```
with requests.get(...) as response:
    response.raise_for_status()
    return response.json()
```

A further fix would be to open a requests session once so that a single connection is used to metabase. I don't know how many tables are involved, but I see a lot of `_get` and `_put` below in loops. It's actually pretty easy to use, see https://requests.readthedocs.io/en/master/user/advanced/#session-objects.

In this case, something like:
```
@property
def session(self):
    if not self._session:
        session = requests.Session()
        # Gather headers
        with session.post(...) as response:
            response.raise_for_status()
            session_key = response.json()['id']
        session.headers.update({
            'Content-Type': 'application/json',
            'X-Metabase-Session': session_key,
        })
        self._session = session
    return self._session
```
Then later using it is easy. You just use `self.session` instead of `requests` and you no longer need to specify the headers because they're set in the session object.
```
def _get(self, path):
    with self.session.get(...) as response:
        response.raise_for_status()
        return response.json()
```

dbnicholson (approved):
This looks good to me, although I didn't look that closely at the actual formatting part. I added a suggestion for optimizing the HTTP usage with metabase, but it can be handled later if you don't want to get into it now. What's there will work fine.

docs/source/metabase.py
liZe: I didn’t know at all that it was possible with Requests… Thanks a lot! I’ll fix the code to use a session.

liZe: > I don't know how many tables are involved, but I see a lot of `_get` and `_put` below in loops.

Indeed. It now takes only ~3 minutes instead of ~15, and the code is cleaner.

dbnicholson: Yay for HTTP connection sharing!

wjt:
Looks like this needs to be refreshed after some parallel changes to the event documentation.

liZe:
> Looks like this needs to be refreshed after some parallel changes to the event documentation.

I’ve included the latest changes of the wiki page.

liZe:
@wjt I’ve rebased this branch onto master, it’s ready to be reviewed and merged.

wjt:
I believe the deployment happens in https://github.com/endlessm/azafea/pull/139, right?

liZe:
> I believe the deployment happens in #139, right?

Yes, in #139.

PR #130, opened 2020-11-09
Merge alembic migrations

liZe:
The alembic tree had 3 heads: e7c7d5f2e4df, 7e12d8f42f28, eba4a3cf4aae. As 3 branches have already been deployed on production, we can’t rebase. We have to merge these heads into one.


adarnimrod (approved):
LGTM. Is this a common situation with Alembic? Can we prevent this from happening again? If so, can you please add some documentation?

liZe:
> If so, can you please add some documentation?

I’ve added documentation. As explained during the latest meeting, having branches is very often harmless, even if it can be confusing. We can try to avoid them when possible, but at least we know what to do if it happens again.

PR #131, opened 2020-11-09
Add index on started_at and occured_at columns

liZe:
**To be merged after #130.**

See https://phabricator.endlessm.com/T29220.

### DIFF
--- a/azafea/event_processors/endless/metrics/events/_base.py
+++ b/azafea/event_processors/endless/metrics/events/_base.py
@@ -182,7 +182,7 @@ class InvalidEvent(UnknownEvent):
 class SingularEvent(MetricEvent):
     __abstract__ = True
 
-    occured_at = Column(DateTime(timezone=True), nullable=False)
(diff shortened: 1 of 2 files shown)


azafea/event_processors/endless/metrics/events/_base.py
```diff
@@ -211,7 +211,7 @@ class UnknownAggregateEvent(AggregateEvent, UnknownEvent):
 class SequenceEvent(MetricEvent):
     __abstract__ = True
 
-    started_at = Column(DateTime(timezone=True), nullable=False)
+    started_at = Column(DateTime(timezone=True), nullable=False, index=True)
```
wjt, line 214: While we're here can we add the same on `occured_at` for other event types?

liZe: Done!

wjt (approved):
LGTM

liZe:
Ready to be merged into `merge-alembic` (into #130, on top of already merged #132).

PR #132, opened 2020-11-23
Add day index on machine by day table

liZe:
See https://phabricator.endlessm.com/T29216


wjt (approved):

wjt:
I see that the target branch is the source branch of #130 so I have just merged it directly there.

PR #134, opened 2020-12-07
Store machine location into metrics_machine table

liZe:
Instead of storing machine location in events, we can store it in the metrics_machine table and upsert when new events happen.

The deployment creates the new columns and sets the values according to the events already stored in the database.

The event table is not removed yet. This way, we can check that everything is OK for a while, and keep events if anything goes wrong.

This is the first step of https://phabricator.endlessm.com/T30471.

### DIFF
--- a/azafea/event_processors/endless/metrics/machine.py
+++ b/azafea/event_processors/endless/metrics/machine.py
@@ -7,16 +7,20 @@
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 
-from typing import Any, Dict
+from typing import Any, Dict, List, Union
+import logging
 
-from sqlalchemy.dialects.postgresql import insert
-from sqlalchemy.schema import Column
+from sqlalchemy.dialects.postgresql import JSONB, insert
+from sqlalchemy.schema import Column, Index
 from sqlalchemy.sql import expression
 from sqlalchemy.types import Boolean, DateTime, Integer, Unicode
 
 from azafea.model import Base, DbSession
 
 from ..image import parse_endless_os_image
+
+
+log = logging.getLogger(__name__)
 
 
 class Machine(Base):
@@ -34,6 +38,28 @@ class Machine(Base):
     demo = Column(Boolean, server_default=expression.false())
     dualboot = Column(Boolean, server_default=expression.false())
     live = Column(Boolean, server_default=expression.false())
+    location = Column(JSONB)
+    location_id = Column(Unicode)
+    location_city = Column(Unicode)
+    location_state = Column(Unicode)
+    location_street = Column(Unicode)
+    location_country = Column(Unicode)
+    location_facility = Column(Unicode)
+
+    __table_args__ = (
+        Index('ix_metrics_machine_location_id', location_id,
+              postgresql_where=(location_id.isnot(None))),
+        Index('ix_metrics_machine_location_city', location_city,
+              postgresql_where=(location_city.isnot(None))),
+        Index('ix_metrics_machine_location_state', location_state,
+              postgresql_where=(location_state.isnot(None))),
+        Index('ix_metrics_machine_location_street', location_street,
+              postgresql_where=(location_street.isnot(None))),
+        Index('ix_metrics_machine_location_country', location_country,
+              postgresql_where=(location_country.isnot(None))),
+        Index('ix_metrics_machine_location_facility', location_facility,
+              postgresql_where=(location_facility.isnot(None))),
+    )
 
 
 def upsert_machine_demo(dbsession: DbSession, machine_id: str) -> None:
@@ -67,3 +93,27 @@ def upsert_machine_live(dbsession: DbSession, machine_id: str) -> None:
                                       set_={'live': True})
 
     dbsession.connection().execute(stmt)
+
+
+def upsert_machine_location(dbsession: DbSession, machine_id: str,
+                            info: Union[Dict[str, Any], List[Any]]) -> None:
+    """Update the relevant Machine record with information from a LocationLabel event.
+
+    Although the info is guaranteed to be a `Dict[str, str]` when received from
+    the client, the database column is of type JSONB, so the
+    `LocationLabel.info` field has a more general type. As a result, the
+    function parameter also has a more general type, and this runtime type
+    check is needed to satisfy mypy.
+
+    """
+    if isinstance(info, dict):
+        location_columns = [column.name for column in Machine.__table__.columns]
+        values = {
+            f'location_{key}': value for key, value in info.items()
+            if f'location_{key}' in location_columns}
+        stmt = insert(Machine.__table__).values(machine_id=machine_id, location=info, **values)
+        stmt = stmt.on_conflict_do_update(constraint='uq_metrics_machine_machine_id', set_=values)
+
+        dbsession.connection().execute(stmt)
+    else:  # pragma: no cover
+        log.warning('Data received for machine location is not a dict: %r', info)
--- a/azafea/event_processors/endless/metrics/v2/migrations/991574509c57_add_machine_location_columns.py
+++ b/azafea/event_processors/endless/metrics/v2/migrations/991574509c57_add_machine_location_columns.py
@@ -1,0 +1,120 @@
+# type: ignore
+
+"""Add machine location columns
+
+Revision ID: 991574509c57
+Revises: 98078d059259
+Create Date: 2020-12-03 12:33:17.219965
+
+"""
+from alembic import op
+from sqlalchemy.dialects.postgresql import JSONB
+from sqlalchemy.sql.expression import text
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = '991574509c57'
+down_revision = '98078d059259'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('metrics_machine', sa.Column('location', JSONB(), nullable=True))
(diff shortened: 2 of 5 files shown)


liZe:
The deployment probably needs to be carefully tested on a dev database before. I’ve launched the request on the production database using a temporary table, I think that it works, but I would be much more confident if the real deployment was tested with production data.

azafea/event_processors/endless/activation/tests/integration/test_activation_v1.py
```diff
@@ -117,4 +117,4 @@ def test_activation_v1_invalid_image(self, capfd):
             assert dbsession.query(Activation).count() == 0
 
         capture = capfd.readouterr()
-        f"Invalid image id 'image': Did not match the expected format" in capture.err
+        assert "Invalid image id 'image': Did not match the expected format" in capture.err
```
wjt, line 120: This change has exposed a pre-existing bug: this assertion fails. (Previously the `is` expression evaluated to `False` which was discarded.)

I suggest separating this patch ("Remove useless f-strings") to a separate PR, unless there's a strong reason to have it in this PR?

azafea/event_processors/endless/metrics/machine.py
```diff
@@ -34,6 +34,12 @@ class Machine(Base):
     demo = Column(Boolean, server_default=expression.false())
     dualboot = Column(Boolean, server_default=expression.false())
     live = Column(Boolean, server_default=expression.false())
+    location_id = Column(Unicode, index=True)
+    location_city = Column(Unicode, index=True)
+    location_state = Column(Unicode, index=True)
+    location_street = Column(Unicode, index=True)
+    location_country = Column(Unicode, index=True)
+    location_facility = Column(Unicode, index=True)
```
wjt, line 42: I think we should store the entire location dictionary here. There is no guarantee which fields it will or will not contain: the client will just blindly submit whatever somebody puts into the configuration file.

Given a JSONB column with the complete dictionary, maybe these columns could be [generated columns](https://www.postgresql.org/docs/12/ddl-generated-columns.html), if our postgresql is new enough for that? … ah no, we have 11.8 and this feature is new in 12. Ah well.

I think it is valuable to have the columns broken out in this way – it makes them much easier to access, whether in SQL (with more familiar syntax) or in Metabase (which doesn't have any special support for JSONB columns in its query builder UI).

I wonder whether these indexes should be [partial](https://www.postgresql.org/docs/8.0/indexes-partial.html) and only index non-empty values. We only have location information for 241 machines, out of 853 739 machines total.

azafea/event_processors/endless/metrics/machine.py
```diff
@@ -67,3 +73,13 @@ def upsert_machine_live(dbsession: DbSession, machine_id: str) -> None:
                                       set_={'live': True})
 
     dbsession.connection().execute(stmt)
+
+
+def upsert_machine_location(dbsession: DbSession, machine_id: str,
+                            info: Union[Dict[str, Any], List[Any]]) -> None:
+    if isinstance(info, dict):
+        info = {f'location_{key}': value for key, value in info.items()}
+        stmt = insert(Machine.__table__).values(machine_id=machine_id, **info)
```
wjt, line 82: As above, there's no guarantee what keys the dictionary will contain. I can put an arbitrary key `foo` into `/etc/metrics/location.conf` on my system, and it will cause this query to fail since there is no `location_foo` column in the table.

azafea/event_processors/endless/metrics/v2/migrations/991574509c57_add_machine_location_columns.py
```diff
@@ -0,0 +1,106 @@
...
+        )
+        ON CONFLICT ON CONSTRAINT
+          uq_metrics_machine_machine_id
+        DO UPDATE SET
+          location_street=excluded.location_street,
+          location_state=excluded.location_state,
+          location_id=excluded.location_id,
+          location_facility=excluded.location_facility,
+          location_country=excluded.location_country,
+          location_city=excluded.location_country;
```
wjt, line 90: This query is a masterpiece!

azafea/event_processors/endless/metrics/machine.py
```diff
@@ -67,3 +73,13 @@ def upsert_machine_live(dbsession: DbSession, machine_id: str) -> None:
                                       set_={'live': True})
 
     dbsession.connection().execute(stmt)
+
+
+def upsert_machine_location(dbsession: DbSession, machine_id: str,
+                            info: Union[Dict[str, Any], List[Any]]) -> None:
+    if isinstance(info, dict):
```
wjt, line 80: Great, looks like inverting the sense of this check did the right thing! Fascinating that mypy can't apply the runtime type check when it's an early return.

I suggest that a comment about this more-general type plus runtime check would be good. Perhaps something like (line wrapped):

```suggestion
    """
    Update the relevant Machine record with information from a LocationLabel event.
    
    Although the info is guaranteed to be a `Dict[str, str]` when received from the client, the database column is of type JSONB, so the `LocationLabel.info` field has a more general type. As a result, the function parameter also has a more general type, and this runtime type check is needed to satisfy mypy.
    """
    if isinstance(info, dict):
```

wjt (changes requested):

azafea/event_processors/endless/metrics/machine.py
liZe: > Great, looks like inverting the sense of this check did the right thing! Fascinating that mypy can't apply the runtime type check when it's an early return.

After reading the previous error again, the problem was not because mypy didn’t understand the `not isinstance` part, but because I had left a comment to ignore types even if they were OK (yes, mypy checks that comments are actually useful oO). I’ve removed the `not` because it was better this way.

> I suggest that a comment about this more-general type plus runtime check would be good. Perhaps something like (line wrapped):

You’re right, that’s a nice gift for the future person who will read this.

azafea/event_processors/endless/activation/tests/integration/test_activation_v1.py
liZe: > I suggest separating this patch ("Remove useless f-strings") to a separate PR, unless there's a strong reason to have it in this PR?

The only reason was that they prevent tests to pass. I don’t know why flake8 suddenly decided that it was bad.

If you want, I can create a PR fixing f-strings that can be merged quickly, and rebase this branch when it’s merged.

azafea/event_processors/endless/metrics/machine.py
liZe: I can add a field storing the JSONB, and keep the new fields with a partial index.

azafea/event_processors/endless/metrics/machine.py
liZe: You’re right, I thought that the keys were fixed.

azafea/event_processors/endless/activation/tests/integration/test_activation_v1.py
wjt: I see, can I suggest 3 separate PRs?

1. The patch to replace redundant `f""` with `""` (but without adding the missing `assert`s)
2. The rest of this PR, built atop 1
3. A patch to add the missing `assert`s you've found and address the issues caught by adding them, also built atop 1

liZe: #137 #138

azafea/event_processors/endless/metrics/machine.py
liZe: Done.

azafea/event_processors/endless/metrics/machine.py
liZe: Done. The dictionary is always updated to the latest version received, and columns are updated for keys included in the received dictionary.

liZe:
> The deployment probably needs to be carefully tested on a dev database before. I’ve launched the request on the production database using a temporary table, I think that it works, but I would be much more confident if the real deployment was tested with production data.

@wjt I’ve updated the SQL request to update `metrics_machine.location` too. It would be great to test the whole deployment on real and manually check the result, before deploying on production.

azafea/event_processors/endless/metrics/machine.py
```diff
@@ -67,3 +93,27 @@ def upsert_machine_live(dbsession: DbSession, machine_id: str) -> None:
...
+        location_columns = [column.name for column in Machine.__table__.columns]
+        values = {
+            f'location_{key}': value for key, value in info.items()
+            if f'location_{key}' in location_columns}
+        stmt = insert(Machine.__table__).values(machine_id=machine_id, location=info, **values)
+        stmt = stmt.on_conflict_do_update(constraint='uq_metrics_machine_machine_id', set_=values)
+
+        dbsession.connection().execute(stmt)
+    else:  # pragma: no cover
+        log.warning('Data received for machine location is not a dict: {%s}', info)
```
wjt, line 119: To me the `{}`s are strongly associated with dictionaries so perhaps just:

```suggestion
        log.warning('Data received for machine location is not a dict: %r', info)
```

wjt (approved):
I only have one very minor comment. I agree that it would be good to test the migration on a dump of the real data!

azafea/event_processors/endless/metrics/machine.py
liZe: Sure.

liZe:
> It would be great to test the whole deployment on real and manually check the result, before deploying on production.

I’ve deployed the migration on a testing database, everything’s fine. :tada:

PR #139, opened 2020-12-23
Use GitHub instead of Travis

liZe:
https://phabricator.endlessm.com/T31180

Secrets have to be added to this repository to allow deployments on Metabase (see #129).

Deployments on ReadTheDocs are possible using a simple webhook, no CI required.

### DIFF
--- a/.github/workflows/metabase.yml
+++ b/.github/workflows/metabase.yml
@@ -1,0 +1,23 @@
+name: Azafea Metabase Documentation
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  metabase:
+    name: Update Documentation on Metabase
+    runs-on: ubuntu-latest
+    env:
+      METABASE_USERNAME: ${{ secrets.METABASE_USERNAME }}
+      METABASE_PASSWORD: ${{ secrets.METABASE_PASSWORD }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install GObject-Introspection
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -0,0 +1,56 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Get Docker Image
+        uses: actions/download-artifact@v2
+        with:
+          name: image
+      - name: Load Docker Image
+        run: docker load < docker-image.tar
+      - name: Install Docker Images
```
adarnimrod, line 35: I think this step could be replaced with [services](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idservices).

.github/workflows/metabase.yml
```diff
@@ -0,0 +1,23 @@
+name: Azafea Metabase Documentation
+on:
+  push:
+    branches:
+      - master
+
+jobs:
```
adarnimrod, line 7: Aren't we missing the update to RTD?

adarnimrod (changes requested):
I had 2 comments. The one about sing services isn't critical, but I think that migrating the RTD update is.

.github/workflows/metabase.yml
wjt: We discussed this verbally – RTD is updated by a web hook.

PR #140, opened 2021-01-07
Don’t use Redis blocking functions

liZe:
The previous code was always triggering a long timeout, even when executed with the `exit_on_empty_queues` option.

Using `sleep()` instead of blocking functions gives code that’s more simple, and that avoids this extra timeout during the first loop.

Another change required by non-blocking functions is that we now use a loop for different queues. Even if the loop adds 3 extra (trivial) lines, it also simplifies code, especially for moking Redis.

This is really useful for tests, as most of them are obviously launched with the `exit_on_empty_queues` option set: it saves 5 seconds for each test relying on Redis queues.

There’s also [another branch](https://github.com/endlessm/azafea/commit/8952debbb38afa5820af38de99068dd2eb4b6899) with a dirty but more simple fix. This PR is probably better.

@wjt This PR looks nice, but I’m afraid to break something just to save 3 minutes for tests. @adarnimrod proposed to try this on a dev server, to check that load and queues are OK. We could also ask Redis and Python gurus to review, if we’re not sure.

### DIFF
--- a/azafea/processor.py
+++ b/azafea/processor.py
@@ -8,6 +8,7 @@
 
 
 import logging
+import time
 from multiprocessing import Process
 from signal import SIGINT, SIGTERM, Signals, signal as intercept_signal
 from typing import Any
@@ -60,7 +61,10 @@ class Processor(Process):
         log.debug('{%s} Pulling from event queues: %s', self.name, queues)
 
         while self._continue:
-            result = self._redis.brpop(queues, timeout=5)
+            for queue in queues:
+                result = self._redis.rpop(queue)
+                if result:
+                    break
 
             if result is None:
                 if self.config.main.exit_on_empty_queues:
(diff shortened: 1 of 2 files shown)


wjt (commented):
I am not a Redis expert but I don't think the new behaviour is quite equivalent to the old.

`BRPOP` will block until an element is pushed onto one of the given queues, or the timeout is reached. This means that if an event is pushed to the queue before that 5-second timeout expires, it will be returned to the event processor immediately.

In the new code, there is an unconditional sleep for 5 seconds before retrying, which will not be interrupted if an element is pushed onto the queue.

Suppose an event is pushed 2 seconds after the call to (b)rpop. Then:

```
Time  0     1     2     3     4     5

Old   BRPOP.......ret

New   RPOPs, sleep .................RPOP ret
```

where "ret" is the point where the processor can process the event.

So the cost of a faster test suite is latency in the production service.

liZe:
> So the cost of a faster test suite is latency in the production service.

It’s true, we get an extra latency when the queue is empty. But that’s only latency, not speed.

The [other branch](https://github.com/endlessm/azafea/commit/8952debbb38afa5820af38de99068dd2eb4b6899) doesn’t have this problem, but we don’t get the simplification we have using `rpop` instead of `brpop`.

A clean (but more complicated) solution would be to use `rpop` when launched once, and `brpop` when launched forever. I’m not sure that it’s worth it.

I can do what you prefer!

liZe:
After discussing this issue yesterday, I finally think that the best solution is just to use a shorter timing when we want to exit on empty queues. No change in the logic, no special case for tests, and code that’s easy to understand.

I’ve uploaded a [third branch](https://github.com/endlessm/azafea/commit/f75b21a47235c469858cd1051da627d26f8c685b) that’s close to the first one, but more readable.

@wjt @adarnimrod If it’s OK for you, I can close this PR and just push this trivial commit to master (or open another PR).

wjt:
Fine by me.

liZe:
Replaced by #141.

PR #141, opened 2021-01-14
Improve tests speed

liZe:
This pull request includes two commits.

1. Use a very short timeout if we want to exit when the Redis queue is empty.
2. Use simple code to check that no table is in database (only used for tests).

### DIFF
--- a/azafea/tests/integration/__init__.py
+++ b/azafea/tests/integration/__init__.py
@@ -16,8 +16,6 @@ import pytest
 
 from redis import Redis
 
-from sqlalchemy.exc import ProgrammingError
-
 import toml
 
 from azafea import cli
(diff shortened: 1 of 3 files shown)


azafea/tests/integration/__init__.py
```diff
@@ -33,17 +31,13 @@ def ensure_tables(self, *models):
                 dbsession.query(model).count()
 
     def ensure_no_tables(self):
+        connection = self.db._engine.connect()
```
wjt, line 34: Can this object be used as a context manager, either directly or with `contextlib.closing()`?

wjt: Amazing that this change has a 10% performance improvement…

liZe: > Can this object be used as a context manager, either directly or with `contextlib.closing()`?

Of course. No idea why I always forget contexts…

> Amazing that this change has a 10% performance improvement…

Creating a connection for each table and for each test was really expensive :wink:.

liZe:
Tests were passing on my computer, but not on GitHub Actions. Why?
The idea of using a short timeout was nice, but it didn’t work. Why?
No error in the logs but no line added in the database, a stupid problem impossible to debug. Why?

Because we use Redis 5, and floats for timeouts are only accepted since … Redis 6.

So, after the third way, here is the fourth way: keep `brpop` for endless loops, and use `rpop` when we want to exit early. This solution is a bit more complicated, but the code is easy to understand, and it’s actually The Good Way™ to avoid a mandatory timeout.

Both cases are actually used in tests: we generally use early exits when we want to test the Redis queue, but we also test signals to stop Azafea running forever. We keep our 100% coverage 🎉.

Bonus: floats are now allowed for timeouts in Redis typing, if we ever want to use them: python/typeshed#4952.

Tests on my computer:
- `master`: ~400s
- first commit: ~240s (65% faster)
- second commit: ~220s (10% faster)

And tests are 2 times faster on GitHub actions.

wjt (approved):

PR #144, opened 2021-02-24
Actions fixup.

adarnimrod:
Can't share artifacts between GitHub Actions workflows, so the job for
publishing the Docker image to Docker Hub needs to be in the tests
workflow (where the image is built).

https://phabricator.endlessm.com/T31467


.github/workflows/tests.yml
```diff
@@ -54,3 +56,30 @@ jobs:
...
       - name: Lint and Type Checking
         run: docker run --rm --network=host --entrypoint="" azafea-tests pipenv run lint
+
+  push:
+    name: Upload the Docker image to Docker Hub
+    runs-on: ubuntu-latest
+    # We only want to publish a new image when the master branch changes, not
+    # for PRs and only if tests finish successfully.
+    needs: tests
+    if: github.event_name == 'push' || github.event_name == 'pull_request'
```
wjt, line 66: Isn't this condition the wrong way round? To me this looks like it will run for pull requests, which is the opposite of what we want

adarnimrod: You're right, but instead of merging to `master` and only then would the job run (and then open another PR with another fix), I added that to the condition and marked the PR as a draft until it's working properly.

adarnimrod:
Publishing the image works now and only runs on pushes to the `master` branch. Ready for review.

.github/workflows/tests.yml
wjt: Ah, I got it!

PR #149, opened 2021-04-29
Add documentation for new wire format

liZe:
https://phabricator.endlessm.com/T31728

### DIFF
--- a/docs/source/events.rst
+++ b/docs/source/events.rst
@@ -11,143 +11,8 @@ versions of Endless OS which are not recorded by current versions.
 Stored Events
 -------------
 
-.. automodule:: azafea.event_processors.endless.metrics.v2.model
+.. automodule:: azafea.event_processors.endless.metrics.v3.model
    :members:
-
-
-Unknown Events
---------------
-
-Desktop Searches
-~~~~~~~~~~~~~~~~
-
-We record searches made from the desktop search bar.
-
-- Since: 2.1.2
-- UUID: ``b02266bc-b010-44b2-ae0f-8f116ffa50eb``
-- UUID name: ``EVENT_DESKTOP_SEARCH`` in gnome-shell
-- Payload: type ``(us)``
-
-  - search provider
-
-    - local: 0
-    - Google: 1
-
-  - query string (what the user searched for)
-
-.. note::
-
-   Since the 2.1.6. release, Google searches have no longer been recorded.
-
-Knowledge App Search
-~~~~~~~~~~~~~~~~~~~~
-
-We record the search terms used for searching within the knowledge apps along
-with the app ID of the knowledge app. (We also record the search term used when
-a user performed a desktop search and clicked through to a knowledge app.)
-
-- Since: 2.3.0
-- UUID: ``a628c936-5d87-434a-a57a-015a0f223838``
-- UUID name: ``SEARCH_METRIC_EVENT_ID`` in eos-knowledge-lib
-- Payload: type ``(ss)``
-
-  - search terms
-  - application ID
-
-Link Shared from Knowledge App
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-Reported when a user shares a link from one of our content apps on a social network.
-
-- Since: SDK 2
-- UUID: ``6775771a-afe7-4158-b7bb-6296fcc7b70d``
-- UUID name: ``SHARE_METRIC_EVENT_ID`` in eos-knowledge-lib
-- Payload: type ``(sayssu)``
-
-  - Application ID (e.g. ``com.endlessm.animals.en``)
-  - ID of content record as a byte array of length 20
-    (``ekn://043fd69fe153ac69a05000b60bfea9cff110f14c`` becomes ``[0x04, 0x3f,
-    0xd6, 0x9f, 0xe1, 0x53, 0xac, 0x69, 0xa0, 0x50, 0x00, 0xb6, 0x0b, 0xfe,
-    0xa9, 0xcf, 0xf1, 0x10, 0xf1, 0x4c]``)
-  - Content title
-  - The exact URL of the content online, as it was shared to the social network
-  - A numerical code indicating which social network the content was shared to
-
-    - Facebook: 0
-    - Twitter: 1
-    - Whatsapp: 2
-
-See `T18524 <https://phabricator.endlessm.com/T18524>`_.
-
-Knowledge App – Article Open/Close
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-We record when an article is opened or closed in a knowledge app. We record the
-ID of the content, the entry point (whether the article was accessed via
-article link, desktop search, app link click, or a nav button), the app ID, the
-article title, and the content type.
-
-- Since: SDK 2
-- UUID: ``fae00ef3-aad7-44ca-aff2-16555e45f0d9``
-- UUID name: ``CONTENT_ACCESS_METRIC_EVENT_ID`` in eos-knowledge-lib
-- Payload of start event: type ``(ssss)``
-
-  - Entry point
-  - Application ID
-  - Title
-  - Content type
-
-- Payload of stop event: ``NULL``
-
-See `T18516 <https://phabricator.endlessm.com/T18516>`_.
-
-Hack Toolbox - Code View Error - Hack Episode 1
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-- Since: 3.5.3
-- UUID: ``e98aa2b8-3f11-4a25-b8e9-b10a635df121``
-- UUID name: ``CODEVIEW_ERROR_EVENT``
-- Payload of start event: type ``sssa(suquq)``
-
-  - Application ID that this toolbox belongs to (string)
-  - ID of function that was being edited in the toolbox (string) (currently always blank)
-  - Contents of code view (string)
-  - List of the error messages that are displayed, each containing:
-  - Text of error message (string)
-  - Start line number where error is shown, 1-based (32-bit unsigned)
-  - Start column number where error is shown, 0-based (16-bit unsigned)
-  - End line number where error is shown, 1-based (32-bit unsigned)
-  - End column number where error is shown, 0-based (16-bit unsigned)
-
-- Payload of progress and stop events: type ``s``
-
-  - Diff of the contents of the code view to the state from the previous event,
-    in the form of an `ed script
-    <https://www.gnu.org/software/diffutils/manual/html_node/ed-Scripts.html>`_,
-    chosen because it's the shortest form that the ``diff`` utility can output
-
-See `T24429 <https://phabricator.endlessm.com/T24429>`_.
-
-Hack Clubhouse - Quest
-~~~~~~~~~~~~~~~~~~~~~~
-
-- Since: 3.7.4
-- UUID: ``50aebb1b-7a93-4caf-8698-3a601a0fc0f6``
-- UUID name: ``QUEST_EVENT``
-- Key: The quest name
-- Payload of start event: type ``(bsas)``
-
-  - ``True`` if the quest is completed
-  - The quest ID
-  - The list of pathway names of this quest
-
-- Payload of stop event: type ``(bsas)``
-
-  - ``True`` if the quest is completed
-  - The quest ID
-  - The list of pathway names of this quest
-
-See `T28004 <https://phabricator.endlessm.com/T28004>`_.
 
 
 Deprecated Events
@@ -414,6 +279,8 @@ See `the comprehensive list of status codes
 Social Bar Is Visible
 ~~~~~~~~~~~~~~~~~~~~~
 
+Removed in TODO.
+
 We record when the social bar is made visible to the user and when it is no
 longer visible. Basically, it corresponds to the user clicking on the social
 bar icon.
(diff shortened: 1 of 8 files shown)


docs/source/protocol.rst
```diff
@@ -30,28 +30,29 @@ Network Send Number
...
-Relative Timestamp
-++++++++++++++++++
+Image
++++++
 
-- 64 bit signed integer
-- GVariant symbol: ``x``
-- See: http://linux.die.net/man/3/clock_gettime
+- String
+- GVariant symbol: ``s``
```
wjt, line 37: We should copy the description & purpose of this string from the `ImageVersion` table and ideally document its structure (from `parse_endless_os_image()`).

docs/source/protocol.rst
```diff
@@ -30,28 +30,29 @@ Network Send Number
...
-Absolute Timestamp
-++++++++++++++++++
+Site
+++++
 
-- 64 bit signed integer
-- GVariant symbol: ``x``
-- Nanoseconds since the Unix epoch
-- See: http://linux.die.net/man/3/clock_gettime
+- String
```
wjt, line 42: Dictionary from string to string

wjt: Could also lift its description from `LocationLabel`

docs/source/protocol.rst
```diff
@@ -30,28 +30,29 @@ Network Send Number
...
+- Boolean
+- GVariant symbol: ``b``
+
+Live
+++++
+
+- Boolean
+- GVariant symbol: ``b``
 
 Singular Metrics
```
wjt, line 57: This section needs update for the new type of each singular metric

docs/source/protocol.rst
```diff
@@ -67,34 +68,19 @@ Aggregate Metrics
 - See `Aggregate Metric`_
 - GVariant symbol: ``a(uayxxmv)``
```
wjt, line 69: Needs update.

docs/source/protocol.rst
```diff
@@ -103,12 +89,17 @@ Event ID
 - Listed in :ref:`events page`
 - See: http://linux.die.net/man/3/uuid
 
-Relative Timestamp
-++++++++++++++++++
+OS Version
+++++++++++
+
+- String
+- GVariant symbol: ``s``
```
wjt, line 96: Could document that this is the `VERSION_ID` from [os-release](https://www.freedesktop.org/software/systemd/man/os-release.html).

docs/source/protocol.rst
```diff
@@ -30,28 +30,29 @@ Network Send Number
 - Begins at 0 on each client and goes up by 1 each time we attempt a metrics network upload.
 - Does not go up on network retries.
 
-Relative Timestamp
```
wjt, line 33: I'm still a little unsure about removing the magic timestamp algorithm…

docs/source/protocol.rst
```diff
@@ -316,3 +240,20 @@ Contents:
 - Singular Events (User ID, Event ID, Relative Timestamp, Auxiliary Payload)
 - Aggregate Events (User ID, Event ID, Count, Relative Timestamp, Auxiliary Payload)
 - Sequence Events (User ID, Event ID, Array of (Relative Timestamp, Auxiliary Payload))
+
+Version 3
+~~~~~~~~~
+
+- Endless X.X.X
```
wjt, line 247:
```suggestion
- Endless 4.0.0
```

wjt (changes requested):

docs/source/protocol.rst
liZe: OK, I changed the `events.rst` page too.

docs/source/protocol.rst
liZe: > Dictionary from string to string

Copy-pasting is bad 👎🏽.

> Could also lift its description from `LocationLabel`

Sure.

docs/source/protocol.rst
liZe: Let’s keep it.

docs/source/protocol.rst
liZe: I’m not sure to understand this comment. Do we need something more than an updated GVariant symbol?

wjt: No, that's all!

docs/source/events.rst
```diff
@@ -11,143 +11,248 @@ versions of Endless OS which are not recorded by current versions.
...
+A user tries to open a ``.exe`` or ``.msi`` file.
+
+:UUID name: ``WINDOWS_APP_OPENED`` in eos-gates
+
+.. versionadded:: 2.1.5
+
+Time Spent In Session
+~~~~~~~~~~~~~~~~~~~~~
+
+Time spent in session by all computer users for given .
```
wjt, line 239:
```suggestion
Time spent in unlocked user session by a particular local user, within the given aggregation period.
```

docs/source/events.rst
```diff
@@ -11,143 +11,248 @@ versions of Endless OS which are not recorded by current versions.
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Duration of an app open during last day.
+
+.. versionadded:: 4.0.0
+
+Different Users
+~~~~~~~~~~~~~~~
+
+Number of different users who have logged in during last week.
```
wjt, line 253:
```suggestion
Number of different users who have logged in during the given aggregation period.
```

docs/source/events.rst
```diff
@@ -11,143 +11,248 @@ versions of Endless OS which are not recorded by current versions.
...
+~~~~~~~~~~~~~~~~~~~~~
+
+Time spent in session by all computer users for given .
+
+.. versionadded:: 4.0.0
+
+Time Foreground App Is Opened
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Duration of an app open during last day.
```
wjt, line 246:
```suggestion
Time in seconds (?? minutes ??) that the given app was open in the foreground of an unlocked session within the given aggregation period.
```

docs/source/events.rst
```diff
@@ -414,6 +519,8 @@ See `the comprehensive list of status codes
 Social Bar Is Visible
 ~~~~~~~~~~~~~~~~~~~~~
 
+Removed in 4.0.0.
```
wjt, line 522: It was removed years ago… 3.2.0 I think.

docs/source/events.rst
```diff
@@ -422,30 +529,809 @@ bar icon.
...
+.. note::
 
-- `Daemon tests <https://github.com/endlessm/eos-event-recorder-daemon/blob/efc6bb0e1283236ee4fe9c3d7fc992c4a53436d8/tests/daemon/test-daemon.c#L51>`_
+   Since the 2.1.6. release, Google searches have no longer been recorded.
 
-  - ``350ac4ff-3026-4c25-9e7e-e8103b4fd5d8``
+Knowledge App Search
+~~~~~~~~~~~~~~~~~~~~
+
+Removed in 4.0.0.
```
wjt, line 558: I'd prefer "not stored since Endless OS 4.0.0". We are unlikely to update the 6 different versions of the Endless SDK (Flatpak) which sends this, but instead just drop it on receipt.

(In fact we never have extracted this event; they're all sitting in the unknown_singular_event table in raw gvariant form.)

docs/source/events.rst
```diff
@@ -422,30 +529,809 @@ bar icon.
...
+- UUID name: ``SEARCH_METRIC_EVENT_ID`` in eos-knowledge-lib
+- Payload: type ``(ss)``
+
+  - search terms
+  - application ID
+
+Link Shared from Knowledge App
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Removed in 4.0.0.
```
wjt, line 575: Ditto above

docs/source/events.rst
```diff
@@ -422,30 +529,809 @@ bar icon.
...
+    - Facebook: 0
+    - Twitter: 1
+    - Whatsapp: 2
+
+See `T18524 <https://phabricator.endlessm.com/T18524>`_.
+
+Knowledge App – Article Open/Close
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Removed in 4.0.0.
```
wjt, line 602: Ditto above

docs/source/events.rst
```diff
@@ -422,30 +529,809 @@ bar icon.
...
+  - Content type
+
+- Payload of stop event: ``NULL``
+
+See `T18516 <https://phabricator.endlessm.com/T18516>`_.
+
+Hack Toolbox - Code View Error - Hack Episode 1
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Removed in 4.0.0.
```
wjt, line 626:
```suggestion
Not reported since Hack was moved to Flathub in Endless OS 3.9.1.
```

docs/source/events.rst
```diff
@@ -422,30 +529,809 @@ bar icon.
...
+    in the form of an `ed script
+    <https://www.gnu.org/software/diffutils/manual/html_node/ed-Scripts.html>`_,
+    chosen because it's the shortest form that the ``diff`` utility can output
+
+See `T24429 <https://phabricator.endlessm.com/T24429>`_.
+
+Hack Clubhouse - Quest
+~~~~~~~~~~~~~~~~~~~~~~
+
+Removed in 4.0.0.
```
wjt, line 655:
```suggestion
Not reported since Hack was moved to Flathub in Endless OS 3.9.1.
```

docs/source/events.rst
```diff
@@ -422,30 +529,809 @@ bar icon.
...
+- ``undefined``: user clicked on a non-clickable area
+
+:UUID name: ``EVENT_DISCOVERY_FEED_CLICK`` in eos-discovery-feed
+
+.. versionadded:: 3.2.0
+
+Discovery Feed Closed
+~~~~~~~~~~~~~~~~~~~~~
+
+Removed in 4.0.0.
```
wjt, line 847: Not reported since 3.9.0 when the Discovery Feed was removed from Endless OS.

docs/source/events.rst
```diff
@@ -422,30 +529,809 @@ bar icon.
...
+popular in and how users generally open the feed.
+
+:UUID name: ``EVENT_DISCOVERY_FEED_OPEN`` in eos-discovery-feed
+
+.. versionadded:: 3.2.0
+
+Disk Space Extra
+~~~~~~~~~~~~~~~~
+
+Removed in 4.0.0.
```
wjt, line 887:
```suggestion
Removed in 4.0.0, along with split-disk support in general.
```

docs/source/events.rst
```diff
@@ -422,30 +529,809 @@ bar icon.
...
+- ``undefined``: user clicked on a non-clickable area
+
+:UUID name: ``EVENT_DISCOVERY_FEED_CLICK`` in eos-discovery-feed
+
+.. versionadded:: 3.2.0
+
+Discovery Feed Opened
+~~~~~~~~~~~~~~~~~~~~~
+
+Removed in 4.0.0.
```
wjt, line 871: Not reported since 3.9.0 when the Discovery Feed was removed from Endless OS.

docs/source/events.rst
```diff
@@ -422,30 +529,809 @@ bar icon.
...
+See `T18983 <https://phabricator.endlessm.com/T18983>`_.
+
+:UUID name: ``DEMO_MODE_ENTERED_METRIC`` in gnome-initial-setup
+
+.. versionadded:: 3.2.5
+
+Endless Application Unmaximized
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Removed in 4.0.0.
```
wjt, line 922: This is another "never going to update the SDK to stop sending this" case – but we can stop storing it.

docs/source/events.rst
```diff
@@ -422,30 +529,809 @@ bar icon.
...
+See `T18445 <https://phabricator.endlessm.com/T18445>`_.
+
+:UUID name: ``EXTRA_DISK_SPACE_EVENT`` in eos-metrics-instrumentation
+
+.. versionadded:: 3.4.3
+
+Entered Demo Mode
+~~~~~~~~~~~~~~~~~
+
+Removed in 4.0.0.
```
wjt, line 902:
```suggestion
Not reported since demo mode was removed in Endless OS 3.9.0
```

docs/source/events.rst
```diff
@@ -422,30 +529,809 @@ bar icon.
...
+We record a metric with the application ID.
+
+:UUID name: ``UNMAXIMIZE_EVENT`` in eos-sdk
+
+.. versionadded:: 3.0.0
+
+Hack Clubhouse Achievement
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Removed in 4.0.0.
```
wjt, line 935:
```suggestion
Not reported since Hack was moved to Flathub in Endless OS 3.9.1.
```

docs/source/events.rst
```diff
@@ -422,30 +529,809 @@ bar icon.
...
+See `T28004 <https://phabricator.endlessm.com/T28004>`_.
+
+:UUID name: ``ACHIEVEMENT_EVENT``
+
+.. versionadded:: 3.7.4
+
+Hack Clubhouse Achievement Points
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Removed in 4.0.0.
```
wjt, line 948:
```suggestion
Not reported since Hack was moved to Flathub in Endless OS 3.9.1.
```

docs/source/events.rst
```diff
@@ -422,30 +529,809 @@ bar icon.
...
+See `T28004 <https://phabricator.endlessm.com/T28004>`_.
+
+:UUID name: ``ACHIEVEMENT_POINTS_EVENT``
+
+.. versionadded:: 3.7.4
+
+Hack Clubhouse Change Page
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Removed in 4.0.0.
```
wjt, line 961:
```suggestion
Not reported since Hack was moved to Flathub in Endless OS 3.9.1.
```

docs/source/events.rst
```diff
@@ -422,30 +529,809 @@ bar icon.
...
+See `T28004 <https://phabricator.endlessm.com/T28004>`_.
+
+:UUID name: ``CLUBHOUSE_SET_PAGE_EVENT``
+
+.. versionadded:: 3.7.4
+
+Hack Clubhouse Enter Pathway
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Removed in 4.0.0.
```
wjt, line 974:
```suggestion
Not reported since Hack was moved to Flathub in Endless OS 3.9.1.
```

docs/source/events.rst
```diff
@@ -422,30 +529,809 @@ bar icon.
...
+See `T28004 <https://phabricator.endlessm.com/T28004>`_.
+
+:UUID name: ``CLUBHOUSE_PATHWAY_ENTER_EVENT``
+
+.. versionadded:: 3.7.4
+
+Hack Clubhouse Mode
+~~~~~~~~~~~~~~~~~~~
+
+Removed in 4.0.0.
```
wjt, line 987:
```suggestion
Not reported since Hack was moved to Flathub in Endless OS 3.9.1.
```

docs/source/events.rst
```diff
@@ -422,30 +529,809 @@ bar icon.
...
+See `T28501 <https://phabricator.endlessm.com/T28501>`_.
+
+:UUID name: ``HACK_MODE_EVENT``
+
+.. versionadded:: 3.7.4
+
+Hack Clubhouse News Quest Link
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Removed in 4.0.0.
```
wjt, line 1000:
```suggestion
Not reported since Hack was moved to Flathub in Endless OS 3.9.1.
```

docs/source/events.rst
```diff
@@ -422,30 +529,809 @@ bar icon.
...
+See `T29192 <https://phabricator.endlessm.com/T29192>`_.
+
+:UUID name: ``CLUBHOUSE_NEWS_QUEST_LINK_EVENT``
+
+.. versionadded:: 3.7.4
+
+Hack Clubhouse Progress
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Removed in 4.0.0.
```
wjt, line 1013:
```suggestion
Not reported since Hack was moved to Flathub in Endless OS 3.9.1.
```

docs/source/events.rst
```diff
@@ -422,30 +529,809 @@ bar icon.
...
+:UUID name: ``MONITOR_DISCONNECTED`` in mutter
+
+UUID was ``ce179909-dacb-4b7e-83a5-690480bf21eb`` before 2.1.7.
+
+.. versionadded:: 2.1.4
+
+ShellAppAddedToDesktop
+~~~~~~~~~~~~~~~~~~~~~~
+
+Removed in 4.0.0.
```
wjt, line 1091:
```suggestion
Removed in 3.9.0.
```

docs/source/events.rst
```diff
@@ -422,30 +529,809 @@ bar icon.
...
+
+    Since 3.9.0 this metric is no longer recorded.
+    See `T30661 <https://phabricator.endlessm.com/T30661>`_.
+
+.. versionadded:: 2.1.0
+
+Shell App Removed From Desktop
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Removed in 4.0.0.
```
wjt, line 1107:
```suggestion
Removed in 3.9.0.
```

docs/source/events.rst
```diff
@@ -422,30 +529,809 @@ bar icon.
...
+
+    Since 3.9.0 this metric is no longer recorded.
+    See `T30661 <https://phabricator.endlessm.com/T30661>`_.
+
+.. versionadded:: 2.1.0
+
+Underscan Enabled
+~~~~~~~~~~~~~~~~~
+
+Removed in 4.0.0.
```
wjt, line 1123:
```suggestion
Removed in 3.9.0.
```

wjt (changes requested):
Mostly looks good. My main quibble with the new wire protocol itself is the use of absolute timestamp in aggregate events.

docs/source/events.rst
liZe: I’ll keep seconds, it’s probably useful for some apps that are open for a short amount of time. I don’t think that it’s particularly bad for privacy.

azafea/event_processors/endless/metrics/events/__init__.py
```diff
@@ -114,6 +114,13 @@ class CacheMetadataIsCorrupt(SingularEvent):
 
 
 class ControlCenterAutomaticUpdates(SingularEvent):
+    """Automatic updates settings have changed.
```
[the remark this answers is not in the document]
liZe: I bet it won’t 💰, because Git mainly relies on blobs, not on paths. I’ll try and move this commit to tables-in-v2 if it conflicts.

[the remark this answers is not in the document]
liZe: No conflict 😎.

docs/source/protocol.rst
```diff
@@ -155,108 +154,44 @@ Event ID
...
+  month)
 
-Relative Timestamp
+Absolute Timestamp
 ++++++++++++++++++
 
+- 64 bit signed integer
+- GVariant symbol: ``x``
+- Nanoseconds since the Unix epoch
+- Beginning of the period, with aggregation done using user’s computer time
```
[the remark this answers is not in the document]
liZe: I don’t know whether I’m right or wrong, but here’s what I have in mind.

Aggregate events need to use the same aggregation period (same start and stop moments) to be useful in database. For example, daily events will be aggregated for everyone between midnight-UTC and midnight-UTC-day+1. This is needed because we want to group rows by this period, to get sums, counts and/or averages by period.

For example:
- (2021-06-07T00:00:00Z, period 'd') represents 2021-06-07 for daily aggregates,
- (2021-06-07T00:00:00Z, period 'w') represents Week 23 of 2021 for weekly aggregates,
- (2021-06-01T00:00:00Z, period 'm') represents June of 2021 for monthly aggregates.

As the client will do the aggregation, it cannot know for sure when the period starts and stops. It has to rely on its own clock to define when the current period stops and the next period starts. Correcting the period (start and stop moments) on the server side using the timestamp algorithm is possible, but it’s useless: we’ll get different "real" periods for each row, losing the ability to group them by period. It’s not technically impossible to use data with different periods, but it breaks the main interest of having aggregates (which is to group by period, in my opinion).

Sequence events don’t have to be grouped by period, that’s why their start/stop times can be shifted on the server.

Having multiple boots is not the problem here. Having the end of the aggregation period wouldn’t be different from having the period.

As a side note, almost fully unrelated: we group information between two moments (between 2021-06-06T00:00:00Z and 2021-06-07T00:00:00Z, with the Z). We may prefer to group information between two naive datetimes (between 2021-06-06T00:00:00 and 2021-06-07T00:00:00, without the Z) to have periods limits when people are supposed to sleep and have clean variations between week days. If we prefer this, it would be better to use dates (local time, just the date) representation in the protocol instead of Epoch (whose values are different for the same local time in two different timezones), and in the database instead of timezone-aware datetimes.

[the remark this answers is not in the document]
wjt: Got it. I think this rationale makes sense. Put another way: we want to be able to summarise across all submitted data by fixed time periods and have that data align reasonably well to those periods. The alignment won't actually be perfect, but close enough is good enough, and we cannot "fix" this on the server-side, as you say.

I'm going back and forth on the naïve versus timezone-aware question… I think it depends on the period. At the level of a week or a month (we should specify whether a week starts on Monday or Sunday…) I don't think it has too big an impact. Perhaps (e.g.) some app usage very early on Monday or late on Sunday is booked against the wrong week; oh well. But at the daily level, I think it may need to be either timezone-aware with machine timezone, or naïve. Otherwise, suppose we have a deployment on the west coast of the USA, expected to be used Monday-Friday (perhaps a school). If we use UTC midnight, that's just 4pm/5pm, and so we would see app/system usage that happens after that point being booked against Saturday. The report would seem "obviously" wrong: the computers are not used on Saturday but metrics say they are.

On the other hand, using local midnight as the cut-off has its own problems: when the timezone changes, the period will not have the correct length. I don't think this is too significant but we'd need to write down the expected behaviour if (e.g.) a computer crosses backwards into a period that has already finished.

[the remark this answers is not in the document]
liZe: As discussed, we’ll use the local timestamp. I’ll fix the code and the documentation on metrics-v3 when this is merged.

liZe:
I’ve rebased `new-protocol-doc` on top of `master` and `metrics-v3`, because the documentation needs the code to be complete. There’s actually no reason to not merge `new-protocol-doc` into `metrics-v3`, before merging `metrics-v3` into master.

I’ll mark this PR as ready for review when `metrics-v3` is rebased on top of `master`.

docs/source/protocol.rst
wjt: This is marked as resolved but it doesn't seem to be resolved.

wjt: This was marked resolved but doesn't seem to be.

docs/source/protocol.rst
wjt: Can't see the description!

docs/source/protocol.rst
```diff
@@ -155,107 +146,41 @@ Event ID
...
-
-Relative Timestamp
-++++++++++++++++++
+Count
++++++
 
 - 64-bit signed integer
-- GVariant symbol ``x``
-- See: http://linux.die.net/man/3/clock_gettime
+- GVariant symbol: ``x``
```
wjt, line 168: Above we document that this is non-negative. So… it should either be a 64-bit **un**signed int. Or, if we are avoiding that type because postgresql only has signed 64-bit ints, we should explain this discrepancy somewhere.

docs/source/protocol.rst
liZe: The commit was on my computer, but not pushed… I’ll reapply the commit on top of the pushed-force new head.

docs/source/protocol.rst
liZe: Idem.

docs/source/protocol.rst
liZe: Right. I’ve transformed this into a 32-bit unsigned integer, it should be big enough and has no problem with PostgreSQL. Downside: it changes the GVariant symbol.

liZe:
I think that it’s ready to be merged. The only change with v3/T32144 is the unsigned 32-bit for aggregate counts, I can add a comment on T32144 when merged. Let me know if this latest change is useless, I can revert it.

I’ve kept the original commits and squashed the various document changes at the end (giving a quite dirty commit). I can rebase more if needed.

wjt:
That seems fine to me. 4 billion seconds (49,710 days) or logged-in users ought to be enough for anyone.

docs/source/protocol.rst
```diff
@@ -45,56 +38,57 @@ Absolute Timestamp
...
+- String
+- Optional human-readable label for the location of the system, which can be
+  used when preparing reports or visualisations of the metrics data
+- GVariant symbol: ``a{ss}``
+
+Flags
++++++
+
+- Byte:
+  - bit 0 is set to 1 for dual-boot
```
wjt, line 61:
```suggestion
  - bit 0 (least-significant bit) is set to 1 for dual-boot
```

docs/source/protocol.rst
wjt: Can we document this?

docs/source/protocol.rst
```diff
@@ -120,33 +119,21 @@ Auxiliary Payload
...
 Aggregate Metric
 ~~~~~~~~~~~~~~~~
 
 Aggregate metrics indicate counts that summarize a value of interest (e.g., a
 very common event happening n times in a particular time interval or
-fluctuations in heap size over time). Counts can be positive, zero, or
-negative. They are identical to the singular metrics but have an added counter
-field in the wire format.
+fluctuations in heap size over time). Counts are always strictly positive. They
+are identical to the singular metrics but have an added counter field in the
```
wjt, line 128: They're not identical to singular metrics any more – occured_at becomes a yyyy-mm-dd interval start.

docs/source/protocol.rst
wjt: Seems to have been lost, still has x.x.x placeholder

wjt (changes requested):

docs/source/protocol.rst
liZe: Right. I’ve added a paragraph about that.

PR #154, opened 2021-06-02
Remove serial and mac_hash

antoinehashbang:
- Remove serial and mac_hash from Activation model
- generate a new migration

https://phabricator.endlessm.com/T31744

### DIFF
--- a/azafea/event_processors/endless/activation/tests/test_activation_parsing.py
+++ b/azafea/event_processors/endless/activation/tests/test_activation_parsing.py
@@ -20,6 +20,8 @@ def test_from_bytes():
         'vendor': 'the vendor',
         'product': 'product',
         'release': 'release',
+        'serial': 'serial',
+        'mac_hash': 694551690,
         'created_at': created_at.strftime('%Y-%m-%d %H:%M:%S.%fZ'),
     }).encode('utf-8'))
 
--- a/azafea/event_processors/endless/activation/v1/handler.py
+++ b/azafea/event_processors/endless/activation/v1/handler.py
@@ -13,7 +13,7 @@ from typing import Optional
 
 from sqlalchemy.orm import validates
 from sqlalchemy.schema import CheckConstraint, Column
-from sqlalchemy.types import BigInteger, Boolean, DateTime, Integer, Numeric, Unicode
+from sqlalchemy.types import Boolean, DateTime, Integer, Numeric, Unicode
 
 from azafea.model import Base, DbSession
 from azafea.vendors import normalize_vendor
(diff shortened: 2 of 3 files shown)


azafea/event_processors/endless/activation/v1/handler.py
```diff
@@ -32,13 +32,9 @@ class Activation(Base):
     vendor = Column(Unicode, nullable=False)
     product = Column(Unicode, nullable=False)
     release = Column(Unicode, nullable=False)
-    serial = Column(Unicode)
```
wjt, line 35: I think we need to make `from_serialized` ignore the `"serial"` and `"mac_hash"` fields by deleting them from `record` before attempting to construct an instance of `Activation`.

That method implicitly assumes that the incoming record only includes keys which are defined as fields on this class. If we were to merge this PR and deploy it, it would fail for all incoming activations because they will still contain `serial` and `mac_hash`. (We can perhaps also make `eos-activation-server` strip them out of incoming requests before putting them into the redis queue.)

I had a quick look at the integration tests and none of them include the `serial` or `mac_hash` fields. Please add an integration test that verifies that these fields are ignored if present in the inccoming request. I think it would be as simple as adding these two fields to the dict in `test_activation_parsing.py`.

wjt (changes requested):

azafea/event_processors/endless/activation/v1/handler.py
antoinehashbang: Ok, I modified `from_serialized` and the test in `test_activation_parsing.` 

But then after testing, sqlalchemy is smart enough to ignore `serial` and `mac_hash` if they are in the record but not in the model fields, I had no error on the tests. But let's remove the fields  from record received anyway!

wjt: That is quite surprising! I would normally expect passing an unexpected kwarg to be fatal. Is that sqlalchemy behaviour documented anywhere?

azafea/event_processors/endless/activation/tests/test_activation_parsing.py
```diff
@@ -20,6 +20,8 @@ def test_from_bytes():
         'vendor': 'the vendor',
         'product': 'product',
         'release': 'release',
+        'serial': 'serial',
+        'mac_hash': 'mac_hash',
```
wjt, line 24:
```suggestion
        'mac_hash': 694551690,
```

This is a more realistic test, being an int not a string.

azafea/event_processors/endless/activation/v1/handler.py
wjt: Or is it specific to the models in Azafea?

If that's so I don't think those fields need to be explicitly dropped. Just the test is fine IMO.

wjt (changes requested):

azafea/event_processors/endless/activation/v1/handler.py
antoinehashbang: Yes, I check, and it's specific to azafea! 
in `azafea.models` on the class `BaseModel` : 
```
def __init__(self, **kwargs: Any):
        columns = inspect(self.__class__).attrs

        for k, v in kwargs.items():
            if k in columns:
                setattr(self, k, v)
```

wjt (approved):

PR #155, opened 2021-06-14
Metrics v3

antoinehashbang:
https://phabricator.endlessm.com/T32086


azafea/event_processors/endless/metrics/v3/migrations/0bd56cdf56b9_initial_migration_metrics_v3.py
```diff
@@ -0,0 +1,857 @@
...
+
+"""
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = "0bd56cdf56b9"
+down_revision = None
+branch_labels = ("metrics-3",)
```
liZe, line 17: Do we really want to have a separate branch for that? I think we do, but I’m not sure 🤔.

azafea/event_processors/endless/metrics/v3/migrations/41a8502c8c17_create_request_model.py
```diff
@@ -0,0 +1,31 @@
+# type: ignore
```
liZe, line 1: We could merge this deployment with the previous one.

azafea/event_processors/endless/metrics/v3/model/__init__.py
```diff
@@ -0,0 +1,625 @@
...
+
+
+# -- Aggregate events ---------------------------------------------------------
+
+class TimeSpentByForegroundApp(AggregateEvent):
+    """Number of seconds spent by an application open in foreground.
+
+    Aggregation is done by day.
+
+    :UUID name: TODO
```
liZe, line 581: Can we define the UUID?

azafea/event_processors/endless/metrics/v3/model/__init__.py
```diff
@@ -0,0 +1,625 @@
...
+            'cpu_info': cpu_info,
+        }
+
+
+# -- Aggregate events ---------------------------------------------------------
+
+class TimeSpentByForegroundApp(AggregateEvent):
+    """Number of seconds spent by an application open in foreground.
+
+    Aggregation is done by day.
```
liZe, line 579: We have to be sure about the aggregation periods.

azafea/event_processors/endless/metrics/v3/model/_base.py
```diff
@@ -0,0 +1,523 @@
...
+
+
+class UnknownSingularEvent(SingularEvent, UnknownEvent):
+    __tablename__ = 'unknown_singular_event_v3'
+
+
+class AggregateEvent(MetricEvent):
+    __abstract__ = True
+
+    period = Column(Unicode(length=1), nullable=False)
```
liZe, line 216: There was a discussion last week about keeping this column. I now think that it’s useless, and that we could use another UUID if we want to change the periodicity.

azafea/event_processors/endless/metrics/v3/model/_base.py
```diff
@@ -0,0 +1,523 @@
...
+
+
+def parse_record(record: bytes) -> Tuple[RequestData, RequestChannel]:
+    # record is an array of bytes, the concatenation of:
+    # - a datetime encoded on 8 bytes (a 64 bits integer timestamp in microseconds)
+    # - the metrics request (a serialized GVariant)
+    # timestamp_bytes, request_bytes = record[:8], record[8:]
+    request_bytes = record[8:]
+
+    # TODO: is this useful?
```
liZe, line 261: I’m not sure about what to do with this timestamp.

azafea/tests/integration/__init__.py
```diff
@@ -119,5 +118,7 @@ def setup_teardown(self, request):
         self.clear_queues()
         self.ensure_no_queues()
 
+        self.db._engine.dispose()
```
liZe, line 121: We discussed about that, it’s strange to get this error now. If the tests are not slower because of that, I think that it’s harmless.

azafea/tests/integration/queries/test_queries.py
```diff
@@ -13,7 +13,7 @@
 class TestQuery(IntegrationTest):
     handler_module = 'azafea.tests.integration.queries.handler_module'
 
-    def test_chunked_query(self):
+    def test_chunked_query_descending(self):
```
liZe, line 16: Why are these tests inverted?

setup.cfg
```diff
@@ -47,7 +47,7 @@ plugins = pydantic.mypy,sqlmypy
 warn_redundant_casts = True
 warn_unused_ignores = True
 
-[mypy-azafea.tests.*,azafea.event_processors.*.tests.*]
+[mypy-azafea.tests.*,azafea.event_processors.*.tests.*,azafea.event_processors.*.tests_v2.*,azafea.event_processors.*.tests_v3.*]
```
liZe, line 50: That was my commit, and it’s a bit ugly. There must be a better way! (© Raymond Hettinger)

azafea/event_processors/endless/metrics/v3/model/_base.py
```diff
@@ -0,0 +1,523 @@
...
+    #: ``state``) to the values provided in the location.conf file (written by
+    #: the ``eos-label-location`` utility)
+    site = Column(JSONB, nullable=False)
+    #: dual boot computer
+    dual_boot = Column(Boolean, nullable=False)
+    #: live USB stick
+    live_usb = Column(Boolean, nullable=False)
+
+
+class Request(Base):
```
liZe, line 87: If the sha is now the same when the data are sent twice, it’s useful to keep this even if it’s not linked to other tables.

liZe (changes requested):
Apart from the comments, it’s OK for me.

We’ll probably have to check/fix little things before merging though:
- Compare the model with the latest version of the model discussed in (T31727)[https://phabricator.endlessm.com/T31727].
- Define UUIDs.
- Reach 100% coverage (21 lines are currently not tested)?
- Build Metabase requests and check that we can get everything we need?
- Test with the new client?

azafea/event_processors/endless/metrics/v3/model/_base.py
wjt: Unfortunately I think it will still change each time because the relative and absolute timestamps will be updated.

azafea/event_processors/endless/metrics/v3/model/__init__.py
```diff
@@ -0,0 +1,625 @@
...
+            'used_disk': payload.get_child_value(2).get_uint32(),
+            'free_disk': payload.get_child_value(3).get_uint32(),
+            'cpu_info': cpu_info,
+        }
+
+
+# -- Aggregate events ---------------------------------------------------------
+
+class TimeSpentByForegroundApp(AggregateEvent):
+    """Number of seconds spent by an application open in foreground.
```
wjt, line 577: There was some discussion in the OS team meeting about whether the right measure may actually be the current one, "time that an app has an open window".

Perhaps we can be less opinionated in the name of this table. `AppUsage{Daily,Monthly}`?

azafea/event_processors/endless/metrics/v3/model/__init__.py
wjt: We discussed yesterday using two fixed aggregation periods, calendar daily and calendar monthly, for these three aggregate events, and hence having 2 × 3 = 6 tables. Each has a unique UUID and the period is a property of that UUID. Is there a good PostgreSQL column type for "calendar day" / "calendar month"?

azafea/event_processors/endless/metrics/v3/model/__init__.py
```diff
@@ -0,0 +1,625 @@
...
+
+    #: application ID
+    app_id = Column(Unicode, nullable=False)
+
+    @staticmethod
+    def _get_fields_from_payload(payload: GLib.Variant) -> Dict[str, Any]:
+        return {'app_id': payload.get_string()}
+
+
+class DifferentUsers(AggregateEvent):
```
wjt, line 598: DailyUsers / MonthlyUsers perhaps a snappier name?

azafea/event_processors/endless/metrics/v3/model/__init__.py
```diff
@@ -0,0 +1,625 @@
...
+
+    .. versionadded:: 4.0.0
+
+    """
+    __tablename__ = 'different_users_v3'
+    __event_uuid__ = 'a3826320-9192-446a-8886-e2129c0ce302'
+    __payload_type__ = None
+
+
+class TimeSpentInSession(AggregateEvent):
```
wjt, line 613: DailySessionTime / MonthlySessionTime

azafea/event_processors/endless/metrics/v3/model/_base.py
```diff
@@ -0,0 +1,523 @@
...
+    #: image ID (e.g. ``eos-eos3.1-amd64-amd64.170115-071322.base``)
+    image_id = Column(Unicode, nullable=False)
+    #: dictionary of string keys (such as ``facility``, ``city`` and
+    #: ``state``) to the values provided in the location.conf file (written by
+    #: the ``eos-label-location`` utility)
+    site = Column(JSONB, nullable=False)
+    #: dual boot computer
+    dual_boot = Column(Boolean, nullable=False)
+    #: live USB stick
+    live_usb = Column(Boolean, nullable=False)
```
wjt, line 84: I would prefer this to just be "live". There are live sessions which are not on USB sticks, such as booting the ISO in a virtual machine.

azafea/event_processors/endless/metrics/v3/model/_base.py
wjt: Isn't it supposed to be combined with the relative + absolute timestamps sent by the client, to calculate how wrong the client's local clock is?

antoinehashbang: In v2, this timestamp was used by Request model, for the field ```received_at``` (who is useless for event creation, it's just an info field about when the request was received) so I think this is useless in v3

antoinehashbang: Except if we want to store when an event is received

setup.cfg
antoinehashbang: We can make it simpler by putting instead ``[mypy-*.tests.*]`` and we should also put the tests_v2 folder in the v2 folder and rename it tests, and same for tests_v3

azafea/event_processors/endless/metrics/v3/model/_base.py
```diff
@@ -0,0 +1,513 @@
...
+    __abstract__ = True
+
+    period_start = Column(Date, nullable=False, index=True)
+    count = Column(BigInteger, nullable=False)
+
+
+class InvalidAggregateEvent(AggregateEvent, InvalidEvent):
+    __tablename__ = 'invalid_aggregate_event_v3'
+
+    receveid_period_start = Column(Unicode)
```
wjt, line 217: "receveid" is a typo – but why is this field needed, can't we use `period_start` from the parent class?

azafea/event_processors/endless/metrics/v3/migrations/0bd56cdf56b9_initial_migration_metrics_v3.py
antoinehashbang: It will be easier to manage the migration files of the two versions and if you want to combine the two branches it will not be very complicated to do

azafea/tests/integration/__init__.py
antoinehashbang: We had no errors before because we never reached the maximum number of connections allowed by postgres. But now that we have the v3 tests, we exceed this number and this causes connection errors to postgres. If I'm not mistaken with this line, at the end of each test we close the connection, which was not done before.

azafea/event_processors/endless/metrics/v3/model/_base.py
```diff
@@ -87,6 +88,24 @@ class Channel(Base):
     live = Column(Boolean, nullable=False)
 
 
+class Request(Base):
+    __tablename__ = 'metrics_request_v3'
```
wjt, line 92: Looks good to me!

azafea/event_processors/endless/metrics/v3/model/_base.py
liZe: We’ll include the timestamp algo in a future PR.

PR #157, opened 2021-06-17
Reduce accuracy GeoIP

antoinehashbang:
- Remove region and city from Activation model
- Generate migrations and add SQL to update existings rows and add
  constraints on latitude and longitude
- add validation on latitude and longitude on Activation model
- Add tests

https://phabricator.endlessm.com/T31730

### DIFF
--- a/azafea/event_processors/endless/activation/v1/handler.py
+++ b/azafea/event_processors/endless/activation/v1/handler.py
@@ -9,6 +9,7 @@
 
 import json
 import logging
+import math
 from typing import Optional
 
 from sqlalchemy.orm import validates
@@ -36,8 +37,6 @@ class Activation(Base):
     live = Column(Boolean)
 
     country = Column(Unicode(length=3))
-    region = Column(Unicode)
-    city = Column(Unicode)
     latitude = Column(Numeric)
     longitude = Column(Numeric)
 
@@ -52,6 +51,8 @@ class Activation(Base):
 
     __table_args__ = (
         CheckConstraint('char_length(country) = 2', name='country_code_2_chars'),
+        CheckConstraint('latitude = floor(latitude) +0.5', name='latitude_precision_reduced'),
+        CheckConstraint('longitude = floor(longitude) +0.5', name='longitude_precision_reduced'),
     )
 
     @validates('country')
@@ -64,6 +65,28 @@ class Activation(Base):
             raise ValueError(f'country has wrong length: {country}')
 
         return country
+
+    @validates('latitude')
+    def validate_latitude(self, key: str, latitude: Optional[int]) -> Optional[int]:
+        # Catch both None and the empty string
+        if not latitude:
+            return None
+
+        if math.floor(latitude) + 0.5 != latitude:
+            raise ValueError(f'latitude is not an integer + 0.5: {latitude}')
+
+        return latitude
+
+    @validates('longitude')
+    def validate_longitude(self, key: str, longitude: Optional[int]) -> Optional[int]:
+        # Catch both None and the empty string
+        if not longitude:
+            return None
+
+        if math.floor(longitude) + 0.5 != longitude:
+            raise ValueError(f'longitude is not an integer + 0.5: {longitude}')
+
+        return longitude
 
     @classmethod
     def from_serialized(cls, serialized: bytes) -> 'Activation':
--- a/azafea/event_processors/endless/activation/v1/migrations/81cf792f49b2_remove_region_city.py
+++ b/azafea/event_processors/endless/activation/v1/migrations/81cf792f49b2_remove_region_city.py
@@ -1,0 +1,45 @@
+# type: ignore
+
+"""Remove region and city and add constraints on latitude and longitude
+
+Revision ID: 81cf792f49b2
+Revises: 3ce72dccdef8
+Create Date: 2021-06-17 12:49:31.415945
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = '81cf792f49b2'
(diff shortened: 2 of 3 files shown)


azafea/event_processors/endless/activation/v1/handler.py
```diff
@@ -52,6 +51,8 @@ class Activation(Base):
 
     __table_args__ = (
         CheckConstraint('char_length(country) = 2', name='country_code_2_chars'),
+        CheckConstraint('latitude = FLOOR(latitude) +0.5', name='latitude_floor'),
+        CheckConstraint('longitude = FLOOR(longitude) +0.5', name='longitude_floor'),
```
wjt, line 55: Perhaps there could be a better name for these constraints. "latitude_precision_reduced"? "latitude_truncated"?

How about verifying they are in the correct range?

azafea/event_processors/endless/activation/v1/handler.py
```diff
@@ -65,6 +66,28 @@ def validate_country(self, key: str, country: Optional[str]) -> Optional[str]:
...
         return country
 
+    @validates('latitude')
+    def validate_latitude(self, key: str, latitude: Optional[int]) -> Optional[int]:
+        # Catch both None and the empty string
+        if not latitude:
+            return None
+
+        if math.floor(latitude) + 0.5 != latitude:
+            raise ValueError(f'latitude has wrong format: {latitude}')
```
wjt, line 76: We can be more specific in this error, perhaps something like:

```suggestion
            raise ValueError(f'latitude is not an integer + 0.5: {latitude}')
```

azafea/event_processors/endless/activation/v1/handler.py
wjt: And let's be consistent in the case of `FLOOR` and `floor` between here and the migration. Lower-case seems conventional according to https://www.postgresql.org/docs/current/functions-math.html:

```suggestion
        CheckConstraint('latitude = floor(latitude) + 0.5', name='latitude_precision_reduced'),
        CheckConstraint('longitude = floor(longitude) + 0.5', name='longitude_precision_reduced'),
```

azafea/event_processors/endless/activation/v1/migrations/81cf792f49b2_remove_region_city.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def upgrade():
+    op.drop_column('activation_v1', 'city')
+    op.drop_column('activation_v1', 'region')
+
+    op.execute(
+        'UPDATE activation_v1 \n'
+        'SET latitude = CASE WHEN latitude=90 THEN 89.5 ELSE floor(latitude)+0.5 END;'
+    )
```
wjt, line 28: In the past when rewriting every row of a table in this way we have sometimes done this as a separate command which processes the table in batches. For example, https://github.com/endlessm/azafea/blob/c146d9b97ef8ead11ff6ae083f9b50de97c6ce3f/azafea/event_processors/endless/ping/v1/cli.py#L97-L124 fills in some generated data in a bunch of columns, and does so in batches of 5000 rows, presumably to avoid holding a lock on the whole table for a long time. There are 1 650 991 rows in this table in the production database. You could use the writable test database to see how long it would take to execute this migration, keeping the table locked. Maybe it is fast enough that we can just do it this simple way.

azafea/event_processors/endless/activation/v1/migrations/81cf792f49b2_remove_region_city.py
```diff
@@ -0,0 +1,48 @@
...
+    op.execute(
+        'UPDATE activation_v1 \n'
+        'SET longitude = CASE WHEN longitude=180 THEN 179.5 ELSE floor(longitude)+0.5 END;'
+    )
```
wjt, line 32: I am not a postgresql expert but I believe that when one updates a row, in fact postgresql writes a new row and marks the old row as garbage. As written, this migration will rewrite the whole table to update latitude, and then immediately rewrite the whole table again to update longitude, generating another 1 650 991 rows of garbage. Is it worth combining the two `UPDATE` statements into one?

wjt (changes requested):

azafea/event_processors/endless/activation/v1/handler.py
wjt: or "latitude has not had its precision reduced"?

dbnicholson:
Looks like the test is failing:
```
self = <azafea.event_processors.endless.activation.v1.handler.Activation object at 0x7fcb0629f850>
key = 'latitude', latitude = 14.3

    @validates('latitude')
    def validate_latitude(self, key: str, latitude: Optional[int]) -> Optional[int]:
        # Catch both None and the empty string
        if not latitude:
            return None
    
        if math.floor(latitude) + 0.5 != latitude:
>           raise ValueError(f'latitude is not an integer + 0.5: {latitude}')
E           ValueError: latitude is not an integer + 0.5: 14.3

azafea/event_processors/endless/activation/v1/handler.py:76: ValueError
```

azafea/event_processors/endless/activation/v1/migrations/81cf792f49b2_remove_region_city.py
antoinehashbang: I tested, if you put both statements in one, it took less than a minute

dbnicholson: Sounds good.

dbnicholson (approved):
Tests are passing and I think this makes sense to do the conversion directly in the migration from one SQL statement since it's fast.

PR #160, opened 2021-07-20
Add request model for metrics v3

antoinehashbang:
https://phabricator.endlessm.com/T32318


azafea/event_processors/endless/metrics/v3/migrations/b0f69709d3f6_add_request_model_metrics_v3.py
```diff
@@ -0,0 +1,64 @@
...
+        sa.UniqueConstraint("sha512", name=op.f("uq_metrics_request_v3_sha512")),
+    )
+    op.create_index(
+        op.f("ix_metrics_request_v3_channel_id"),
+        "metrics_request_v3",
+        ["channel_id"],
+        unique=False,
+    )
+    op.drop_column("invalid_aggregate_event_v3", "receveid_period_start")
+    op.drop_column("unknown_aggregate_event_v3", "receveid_period_start")
```
wjt, line 45: Where did these columns come from? Why are they being removed in this migration? It looks like they were inadvertently present in the migrations which added these tables.

Assuming that's right: this fix should be in a separate migration, added in a separate commit, with a meaningful commit message.

wjt (changes requested):

azafea/event_processors/endless/metrics/v3/migrations/b0f69709d3f6_add_request_model_metrics_v3.py
antoinehashbang: Yes, I saw this today, it was a column I had added and then removed in this table and I must have forgotten to re-generate the migrations after I removed them.

I'm going to put this in a separate commit, would you prefer I leave it in this PR or open another one for it?

wjt:
This PR is fine!

azafea/event_processors/endless/metrics/v3/migrations/112b12fa4dc5_remove_column_receveid_period_start.py
```diff
@@ -0,0 +1,34 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '112b12fa4dc5'
+down_revision = 'a82a26763a82'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.drop_column('invalid_aggregate_event_v3', 'receveid_period_start')
```
wjt, line 22: Please explain in the commit message where these columns came from and why they should be removed.

antoinehashbang: I've updated the commit message!

wjt (approved):

PR #161, opened 2021-07-20
Don’t crash when no table needs a Metabase documentation update

liZe:
See https://phabricator.endlessm.com/T32311.

### DIFF
--- a/docs/source/metabase.py
+++ b/docs/source/metabase.py
@@ -114,6 +114,7 @@ class MetabaseBuilder(DummyBuilder):
         # Tables and database ID
         logger.info('\nUpdating tables\n')
         tables = [table for table in self._get('table') if table['db']['name'] == 'Azafea']
+        db_id = None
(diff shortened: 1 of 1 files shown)


docs/source/metabase.py
```diff
@@ -132,6 +133,10 @@ def finish(self) -> None:
             self._put(f'table/{table["id"]}', json={
                 'description': description, 'points_of_interest': points_of_interest})
 
+        # Quit early if no table has been found
```
wjt, line 136: but why are no tables found? the v2 tables still exist, don't they?

liZe: They’re not in the Sphinx documentation anymore, that’s why they’re not available here.

wjt: Oh… I guess this makes sense. And the existing documentation is already in Metabase so it's fine to not update it.

PR #162, opened 2021-07-22
Use the server’s clock when it’s obviously better than the client’s

liZe:
This is the last part of the timestamp algorithm:
https://azafea.readthedocs.io/en/latest/timestamp-algorithm.html#server-side

"When the server receives a network request, it will first examine the network request absolute time to see if it varies significantly from its own (trusted) clock. If it does, some special action will be taken with that request, such as putting it in its own special table or attempting to correct the timestamp in some fashion."

Our "attempt to correct the timestamp" happens when:
- the request is received before it has been sent (1 second tolerance),
- the request is received a long time after it has been sent (10 minutes tolerance).
In these two cases, we prefer to use the server’s clock, because the client’s clock is obviously wrong.

~https://phabricator.endlessm.com/T29275~ https://phabricator.endlessm.com/T32195

### DIFF
--- a/azafea/event_processors/endless/metrics/tests_v3/test_request.py
+++ b/azafea/event_processors/endless/metrics/tests_v3/test_request.py
@@ -69,3 +69,40 @@ def test_request_builder_invalid():
 
     assert ('Metric request is not in the expected format: '
             '(xxsa{ss}ya(aysxmv)a(ayssumv))') in str(excinfo.value)
+
+
+def test_request_builder_wrong_client_clock():
+    from azafea.event_processors.endless.metrics.v3.model import parse_record
+
+    now = datetime.now(tz=timezone.utc)
+    request = GLib.Variant(
+        '(xxsa{ss}ya(aysxmv)a(ayssumv))',
+        (
+            2000000,   # request relative timestamp (2 secs)
+            int((now - timedelta(hours=1)).timestamp() * 1000000000),  # Abs. timestamp, 1 hour late
+            'image_id',
+            {},
+            2,
+            [],                                    # singular events
+            []                                     # sequence events
+        )
+    )
+    assert request.is_normal_form()
+    request_body = request.get_data_as_bytes().get_data()
+
+    received_at = now + timedelta(minutes=2)
+    received_at_timestamp = int(received_at.timestamp() * 1000000)  # timestamp as microseconds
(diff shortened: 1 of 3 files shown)


wjt:
Wrong ticket, I think https://phabricator.endlessm.com/T32195 is the one, and the linked one can track the existence of the problem for existing v2 data.

@dbnicholson could you take a look here?

liZe:
It fixes ~10% of the events we receive. For the record, interesting stats are here: https://phabricator.endlessm.com/T31727#884096.

azafea/event_processors/endless/metrics/tests_v3/test_request.py
```diff
@@ -69,3 +69,40 @@ def test_request_builder_invalid():
...
+
+def test_request_builder_wrong_client_clock():
+    from azafea.event_processors.endless.metrics.v3.model import parse_record
+
+    now = datetime.now(tz=timezone.utc)
+    request = GLib.Variant(
+        '(xxsa{ss}ya(aysxmv)a(ayssumv))',
+        (
+            2000000,   # request relative timestamp (2 secs)
+            int(now.timestamp() - 3600 * 1000000000),  # Absolute timestamp, 1 hour late
```
dbnicholson, line 82: Is this right? On my local system this gives:
```
>>> int(now.timestamp() - 3600 * 1000000000)
-3598373029762
```
Maybe you meant `int((now.timestamp() - 3600) * 1000000000)`? I think it would be clearer if you used `now - timedelta(hours=1)`. I'm not familiar with how the timestamps are used in azafea, though.

dbnicholson (commented):
This seems good, but I had a question about the test.

azafea/event_processors/endless/metrics/tests_v3/test_request.py
liZe: > Maybe you meant `int((now.timestamp() - 3600) * 1000000000)`? I think it would be clearer if you used `now - timedelta(hours=1)`.

Indeed.

dbnicholson (approved):
Looks good. I don't know how the test was passing before, but this makes more sense.

liZe:
> Looks good. I don't know how the test was passing before, but this makes more sense.

When the difference is big, we keep the server time, no matter how big the difference is. The difference was huge with the previous code, it’s now just an hour, but the result is the same and the test passes in both cases.

PR #167, opened 2021-11-04
Pin dependencies for Read the Docs

dbnicholson:
Rather than duplicate the Read the Docs requirements in a separate file,
include any needed there in the Pipfile dev-packages and generate the
Read the Docs requirements file from that. This will ensure that Read
the Docs builds the documentation using the same versions of the tools
as our own tests.

This is actually working around an issue with the default dependencies
that Read the Docs will install if they're unspecified. In the previous
configuration, `sphinx` was unspecified in
`requirements.readthedocs.txt` and RTD installs 1.8.5 for this project.
See https://docs.readthedocs.io/en/stable/builds.html#python for
details. That version is now broken with a recent `docutils` update
that's incompatible with `sphinx`. A newer `sphinx` is needed where the
`docutils` version is pinned.

https://phabricator.endlessm.com/T32700


CONTRIBUTING.md
```diff
@@ -112,6 +112,23 @@ Both are run automatically with the following command:
...
+run `pipenv lock` to update the lock file. The `--keep-outdated` option
+can be used to keep existing pins from being updated.
+
+The documentation is also built on [Read the
+Docs](https://readthedocs.org/), but that service doesn't support
+Pipenv. Dependencies for Read the Docs are managed in the
+[requirements.readthedocs.txt](requirements.readthedocs.txt) file, which
+is generated from `Pipfile.lock` by running `pipenv lock --requirements
+--dev > requirements.readthedocs.txt`. Please update this file when
+updating `Pipfile.lock`.
```
wjt, line 129: I'm not quite sure where in the GitHub Actions workflow to put this, but: presumably this output is deterministic, since that's kind of the point of locking the dependencies. So we can check in CI that:

```
diff -u requirements.readthedocs.txt <(pipenv lock --requirements --dev)
```

exits with status 0, or I guess,

```
pipenv lock --requirements --dev > requirements.readthedocs.txt
git diff --exit-code requirements.readthedocs.txt
```

wjt (approved):
LGTM but if you are feeling enthusiastic you may wish to enforce the documented requirement to update the rtd requirements in CI.

CONTRIBUTING.md
dbnicholson: I meant to say this in the commit message and forgot. I explicitly ran this with `--keep-outdated` since I didn't want to pull in a bunch of updates randomly. That said, we should probably do that going forward. I just didn't want to roll that into the same work. I'm going to punt on that for now but I opened #168 to handle that in the future.

dbnicholson: Oh, I think I focused on the wrong thing there. You just want to make sure that someone did actually updated `requirements.readthedocs.txt` in CI. I think that can be done. Let me try to put that in.

dbnicholson: I added something now that seems to work.

PR #169, opened 2021-11-18
Fix Read the Docs dependency handling

dbnicholson:
Unfortunately, we can't install in all the dev packages in the Read the
Docs build environment since some of them require system packages.
Instead, just include the sphinx related dependencies on top of the main
requirements. A script is provided to generate the requirements file
consistently and check that it's in sync with `Pipfile.lock`.

https://phabricator.endlessm.com/T32700


wjt (approved):
This looks plausible, and I like that the actions workflow will check that the requirements file is up to date and emit a diff if it is not.

dbnicholson:
Oh, you already reviewed. Merging...

PR #171, opened 2021-12-03
metrics: Add SplitFlatpakRepoStats event

dbnicholson:
This metric is sent by the `eos-split-flatpak-repo` script to record
statistics about splitting the system Flatpak repo from the system
OSTree repo.

The corresponding client side is in endlessm/eos-boot-helper#361.

https://phabricator.endlessm.com/T32768


dbnicholson:
Wrong task number...

PR #172, opened 2021-12-03
metrics: Add SplitFlatpakRepoStats event

dbnicholson:
This metric is sent by the `eos-split-flatpak-repo` script to record
statistics about splitting the system Flatpak repo from the system
OSTree repo.

The corresponding client side is in endlessm/eos-boot-helper#361.

https://phabricator.endlessm.com/T32790

### DIFF
--- a/azafea/event_processors/endless/metrics/tests_v3/integration/test_metrics_v3.py
+++ b/azafea/event_processors/endless/metrics/tests_v3/integration/test_metrics_v3.py
@@ -187,7 +187,8 @@ class TestMetrics(IntegrationTest):
             LaunchedExistingFlatpak, LaunchedInstallerForFlatpak, LinuxPackageOpened,
             ParentalControlsBlockedFlatpakInstall, ParentalControlsBlockedFlatpakRun,
             ProgramDumpedCore, UpdaterFailure, ParentalControlsEnabled,
-            ParentalControlsChanged, WindowsAppOpened, ComputerInformation
+            ParentalControlsChanged, WindowsAppOpened, ComputerInformation,
+            SplitFlatpakRepoStats,
         )
         from azafea.event_processors.endless.metrics.v3.model import _base
         _base.IGNORED_EMPTY_PAYLOAD_ERRORS = ['9d03daad-f1ed-41a8-bc5a-6b532c075832']
@@ -200,7 +201,8 @@ class TestMetrics(IntegrationTest):
             LaunchedExistingFlatpak, LaunchedInstallerForFlatpak, LinuxPackageOpened,
             ParentalControlsBlockedFlatpakInstall, ParentalControlsBlockedFlatpakRun,
             ProgramDumpedCore, UpdaterFailure, ParentalControlsEnabled,
-            ParentalControlsChanged, WindowsAppOpened, ComputerInformation
+            ParentalControlsChanged, WindowsAppOpened, ComputerInformation,
+            SplitFlatpakRepoStats,
         )
 
         # Build a request as it would have been sent to us
@@ -370,7 +372,27 @@ class TestMetrics(IntegrationTest):
                                 [('model_1', 8, 14.5), ('model_2', 8, 14.5)]
                             )
                         )
-                    )
+                    ),
+                    (
+                        UUID('880fd091-2e68-4bd2-baa1-f6810fabcc1c').bytes,
+                        'os_version',
+                        100,
+                        GLib.Variant(
+                            '(uuuuuuutut)',
+                            (
+                                700,
+                                10,
+                                500,
+                                0,
+                                1000000,
+                                0,
+                                200,
+                                10000000000,
+                                50,
+                                20000000000,
+                            )
+                        )
+                    ),
                 ],
                 [],                       # aggregate events
             )
@@ -503,6 +525,19 @@ class TestMetrics(IntegrationTest):
 
             assert dbsession.query(InvalidSingularEvent).count() == 0
             assert dbsession.query(UnknownSingularEvent).count() == 0
+
+            split_flatpak_repo_stats = dbsession.query(SplitFlatpakRepoStats).one()
+            assert split_flatpak_repo_stats.os_version == 'os_version'
+            assert split_flatpak_repo_stats.elapsed == timedelta(seconds=700)
+            assert split_flatpak_repo_stats.num_os_refs == 10
+            assert split_flatpak_repo_stats.num_flatpak_refs == 500
+            assert split_flatpak_repo_stats.num_other_refs == 0
+            assert split_flatpak_repo_stats.num_objects == 1000000
+            assert split_flatpak_repo_stats.num_deltas == 0
+            assert split_flatpak_repo_stats.num_apps == 200
+            assert split_flatpak_repo_stats.size_apps == 10000000000
+            assert split_flatpak_repo_stats.num_runtimes == 50
+            assert split_flatpak_repo_stats.size_runtimes == 20000000000
 
     def test_aggregate_events(self):
         from azafea.event_processors.endless.metrics.v3.model import (
@@ -644,7 +679,8 @@ class TestMetrics(IntegrationTest):
             ParentalControlsBlockedFlatpakInstall, ParentalControlsBlockedFlatpakRun,
             ProgramDumpedCore, UpdaterFailure, ParentalControlsEnabled,
             ParentalControlsChanged, WindowsAppOpened, DailyAppUsage, MonthlyAppUsage,
-            DailyUsers, MonthlyUsers, DailySessionTime, MonthlySessionTime
+            DailyUsers, MonthlyUsers, DailySessionTime, MonthlySessionTime,
+            SplitFlatpakRepoStats,
         )
         self.run_subcommand('initdb')
         self.ensure_tables(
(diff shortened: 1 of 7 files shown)


azafea/event_processors/endless/metrics/tests_v3/integration/test_metrics_v3.py
```diff
@@ -187,7 +187,8 @@ def test_singular_events(self):
             LaunchedExistingFlatpak, LaunchedInstallerForFlatpak, LinuxPackageOpened,
             ParentalControlsBlockedFlatpakInstall, ParentalControlsBlockedFlatpakRun,
             ProgramDumpedCore, UpdaterFailure, ParentalControlsEnabled,
-            ParentalControlsChanged, WindowsAppOpened, ComputerInformation
+            ParentalControlsChanged, WindowsAppOpened, ComputerInformation,
+            SplitFlatpakRepoStats,
```
wjt, line 191: Thanks for adding the trailing comma. I think that in a different life it would be nice for these lists to be one-per-line and sorted alphabetically.

azafea/event_processors/endless/metrics/v3/model/__init__.py
```diff
@@ -570,6 +570,39 @@ def _get_fields_from_payload(payload: GLib.Variant) -> Dict[str, Any]:
         }
 
 
+class SplitFlatpakRepoStats(SingularEvent):
+    """Statictics for splitting the Flatpak repo from the OSTree repo.
```
wjt, line 574:
```suggestion
    """Statistics for splitting the Flatpak repo from the OSTree repo.
```

azafea/event_processors/endless/metrics/v3/model/__init__.py
```diff
@@ -570,6 +570,39 @@ def _get_fields_from_payload(payload: GLib.Variant) -> Dict[str, Any]:
...
+
+    Sent at most once.
```
wjt, line 576: Let's give a bit more detail. I suggest (but with line wraps):

```suggestion
    Prior to Endless OS 4, both the OS and Flatpaks were stored in a single, shared OSTree repository. On the first boot into Endless OS 4 or later, these are split into two repositories, one for the OS and one for Flatpaks. This can take a long time, and blocks the boot process, so we report this event to quantify how long it takes and why.
    
    This is a one-time operation which only runs if needed, so this event is sent at most once per Endless OS system.
```

wjt (changes requested):
Obviously this is contingent on any changes to the reported data over in https://github.com/endlessm/eos-boot-helper/pull/361

azafea/event_processors/endless/metrics/tests_v3/integration/test_metrics_v3.py
dbnicholson: Indeed, for another day.

azafea/event_processors/endless/metrics/v3/model/__init__.py
```diff
@@ -570,6 +570,64 @@ def _get_fields_from_payload(payload: GLib.Variant) -> Dict[str, Any]:
...
+            'elapsed': payload.get_child_value(0).get_uint32(),
+            'num_os_refs': payload.get_child_value(1).get_uint32(),
+            'num_flatpak_refs': payload.get_child_value(2).get_uint32(),
+            'num_other_refs': payload.get_child_value(3).get_uint32(),
+            'num_objects': payload.get_child_value(4).get_uint32(),
+            'num_deltas': payload.get_child_value(5).get_uint32(),
+            'num_apps': payload.get_child_value(6).get_uint32(),
+            'size_apps': payload.get_child_value(7).get_uint64(),
+            'num_runtimes': payload.get_child_value(8).get_uint32(),
+            'size_runtimes': payload.get_child_value(9).get_uint64(),
```
wjt, line 627: PostgreSQL doesn't support 64-bit unsigned integers. The largest type it supports is 64-bit signed int. https://www.postgresql.org/docs/11/datatype-numeric.html

If you try to insert an out-of-range number, the `INSERT` will just fail:

```
azafea=> create temporary table temp_foo (my_large_number bigint);
CREATE TABLE
azafea=> insert into temp_foo values (18446744073709551615); -- 2 ** 64 - 1
ERROR:  bigint out of range
```

There is a `clamp_to_int64` function to work around this. It logs an error if the value is out of range. But the error message is hard to interpret because there is no context provided about where the number came from. Here's an example from searching the logs in Cloudfront:

```
[ERROR] azafea.event_processors.endless.metrics.v2.utils: Clamped integer larger than MAXINT64: 13752137942398757368
```

So at a minimum you want:

```suggestion
            'size_apps': clamp_to_int64(payload.get_child_value(7).get_uint64()),
            'num_runtimes': payload.get_child_value(8).get_uint32(),
            'size_runtimes': clamp_to_int64(payload.get_child_value(9).get_uint64()),
```

but if you were feeling keen you could add a second string parameter to indicate where the number came from.

Another option would be to make these values signed in the wire protocol. On principle this feels wrong because (a) `flatpak_installed_ref_get_installed_size` returns uint64; (b) a size cannot be negative. But it's an option!

azafea/event_processors/endless/metrics/v3/model/__init__.py
```diff
@@ -570,6 +570,64 @@ def _get_fields_from_payload(payload: GLib.Variant) -> Dict[str, Any]:
...
 
 
+class SplitFlatpakRepoStats(SingularEvent):
+    """Statistics for splitting the Flatpak repo from the OSTree repo.
+
+    Prior to Endless OS 4, both the OS and Flatpaks were stored in a
+    single, shared OSTree repository. On the first boot into Endless OS
+    4 or later, these are split into two repositories, one for the OS
+    and one for Flatpaks. This can take a long time, and blocks the boot
+    process, so we report this event to quantify how long it takes and
```
wjt, line 580:
```suggestion
    and one for Flatpaks. This can take a long time,
    so we report this event to quantify how long it takes and
```

azafea/event_processors/endless/metrics/v3/model/__init__.py
```diff
@@ -570,6 +570,64 @@ def _get_fields_from_payload(payload: GLib.Variant) -> Dict[str, Any]:
...
+    :UUID name: ``SPLIT_FLATPAK_REPO_STATS`` in eos-boot-helper
+
+    .. versionadded:: 4.0.1
+    """
+    __tablename__ = 'split_flatpak_repo_stats_v3'
+    __event_uuid__ = '880fd091-2e68-4bd2-baa1-f6810fabcc1c'
+    __payload_type__ = '(uuuuuuutut)'
+
+    #: time elapsed in seconds
+    elapsed = Column(Integer, nullable=False)
```
wjt, line 595: If you were feeling fancy you could store this as an `interval` instead. [sqlalchemy](https://docs.sqlalchemy.org/en/14/core/type_basics.html#sqlalchemy.types.Interval) maps this to `datetime.timedelta` and an `Interval` column type.

This would be nice because it allows the `psql` command line (and Metabase) to format the duration nicely:

```
azafea=> create temporary table blahblah (number int, int interval);
CREATE TABLE
azafea=> insert into blahblah values (86400, '86400 seconds');
INSERT 0 1
azafea=> select * from blahblah;
 number |   int    
--------+----------
  86400 | 24:00:00
(1 row)
```

azafea/event_processors/endless/metrics/v3/model/__init__.py
```diff
@@ -570,6 +570,64 @@ def _get_fields_from_payload(payload: GLib.Variant) -> Dict[str, Any]:
...
+    size_apps = Column(BigInteger, nullable=False)
+    #: number of Flatpak runtimes duplicated
+    num_runtimes = Column(Integer, nullable=False)
+    #: size of Flatpak runtimes duplicated in bytes
+    size_runtimes = Column(BigInteger, nullable=False)
+
+    @staticmethod
+    def _get_fields_from_payload(payload: GLib.Variant) -> Dict[str, Any]:
+        return {
+            'elapsed': payload.get_child_value(0).get_uint32(),
```
wjt, line 618: If you were to change the type of this field to `Interval` you would want this change I think:

```suggestion
            'elapsed': datetime.timedelta(seconds=payload.get_child_value(0).get_uint32()),
```

azafea/event_processors/endless/metrics/v3/model/__init__.py
dbnicholson: Ah, the dreaded exabyte flatpak. Since the issue is here, I think it makes more sense to send the proper data over the wire and handle the issue in azafea.

To fixup `clamp_to_int64`, I'll just get `logging` to print the traceback rather than change all the callers:
```
>>> try:
...     raise ValueError('fail')
... except ValueError:
...     logging.exception('i guess i can let it go')
... 
ERROR:root:i guess i can let it go
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
ValueError: fail
```

azafea/event_processors/endless/metrics/v3/model/__init__.py
dbnicholson: OK, I'll take a look.

wjt (approved):
Ah yes, shame about that duplication...

PR #173, opened 2021-12-09
metrics: Add migration for SplitFlatpakRepoStats event table

dbnicholson:
https://phabricator.endlessm.com/T32790


wjt:
Pushed a change to remove boilerplate comments and merged.

PR #178, opened 2023-05-12
tests: Stop using future-deprecated datetime.utcnow()

wjt:
https://blog.ganssle.io/articles/2019/11/utcnow.html lays out why this function is bad. In fact I believe the usage in the tests is safe. However Python 3.12 will formally deprecate this function: https://discuss.python.org/t/deprecating-utcnow-and-utcfromtimestamp/26221

Replace it with the recommended (and shorter!) alternative.


dbnicholson (approved):
Nearly a year later, I do agree with this and think it's should provide no change in behavior since the timezone was being applied immediately.

PR #179, opened 2023-11-24
Drop tour response event

wjt:
As described in the main commit, we have learned what we wanted to learn by collecting this event. We can now remove the data from the database and stop storing new submissions.

https://phabricator.endlessm.com/T35061


dbnicholson (approved):
I see the documentation build is failing due to changes in `pipenv`. My experience with that tool is mixed at best.

Anyways, that has nothing to do with the changes here and we can fix that another day.

PR #181, opened 2024-03-22
Add CheckpointBlocked{Daily,Monthly} events

wjt:
I chose not to add a _v3 suffix to the table names because:

a. It is incredibly ugly
b. I hope that in time we can retire the old tables

https://phabricator.endlessm.com/T35250

### DIFF
--- a/azafea/event_processors/endless/metrics/v3/model/__init__.py
+++ b/azafea/event_processors/endless/metrics/v3/model/__init__.py
@@ -16,6 +16,7 @@ from sqlalchemy.dialects.postgresql import JSONB
 from sqlalchemy.schema import Column, Index
 from sqlalchemy.types import ARRAY, BigInteger, Boolean, Integer, Interval, Unicode
 
+from azafea.vendors import normalize_vendor
 from ..utils import aggregate_uuid, clamp_to_int64, get_child_values
 from ._base import (  # noqa: F401
     AGGREGATE_EVENT_MODELS,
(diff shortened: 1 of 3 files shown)


azafea/event_processors/endless/metrics/v3/model/__init__.py
```diff
@@ -743,3 +744,82 @@ class MonthlySessionTime(AggregateEvent):
...
+    def _get_fields_from_payload(payload: GLib.Variant) -> Dict[str, Any]:
+        return {
+            "vendor": normalize_vendor(payload.get_child_value(0).get_string()),
+            "product": payload.get_child_value(1).get_string(),
+            "booted_ref": payload.get_child_value(2).get_string(),
+            "target_ref": payload.get_child_value(3).get_string(),
+            "reason": payload.get_child_value(4).get_string(),
+        }
+
+class CheckpointBlockedMonthly(AggregateEvent):
```
wjt, line 792: It's really annoying that this class is a copy-pasta of CheckpointBlockedDaily.

It can't be a subclass of it because that has special meeting in sqlalchemy: the subclass is a separate table with a foreign key to its parent class's table.

It could be a separate abstract base class that does not subclass AggregateEvent, and then this and the Daily table could multiple-inherit from the ABC and from AggregateEvent. But at a certain point, it's not worth it.

wjt:
The *Documentation Building Test* failure is also the case on `master`, not introduced by this branch.

azafea/event_processors/endless/metrics/v3/model/__init__.py
dbnicholson: That is unfortunate, but I agree not worth it at this time.

dbnicholson (approved):
Looks legit. I agree that pre-emptively suffixing the version onto the table name is unnecessary.

dbnicholson:
> The _Documentation Building Test_ failure is also the case on `master`, not introduced by this branch.

I'm gonna take a whack at fixing it since I was the one that introduced the thing that's broken now.

PR #183, opened 2024-03-22
Discard MonitorConnected and MonitorDisconnected

wjt:
We do not have a use for this data.

https://phabricator.endlessm.com/T33684


dbnicholson (approved):
This looks good. I'm not going to merge as I recall that the revisions in the alembic migration may need to be adjusted. Merge away if that's not the case.

wjt:
The revision doesn't need to be revised because:

1. We have one Alembic branch per event processor; the change in #181 was on the v3 event processor but this is on v2
2. AFAICT there is no practical problem in having divergent migration branches within one event-processor; it's more one of tidiness.

wjt:
However I do need to fix:

```
=================================== FAILURES ===================================
_________________________________ FLAKE8-check _________________________________
/opt/azafea/src/azafea/event_processors/endless/metrics/v2/model/__init__.py:19:1: F401 'sqlalchemy.types.LargeBinary' imported but unused
/opt/azafea/src/azafea/event_processors/endless/metrics/v2/model/__init__.py:32:1: F401 '..utils.get_bytes' imported but unused
```

PR #185, opened 2024-03-25
Discard various data that we no longer need

wjt:
See each individual commit message for explanations.

https://phabricator.endlessm.com/T33684


azafea/event_processors/endless/metrics/v3/model/_base.py
```diff
@@ -43,6 +43,7 @@
     '350ac4ff-3026-4c25-9e7e-e8103b4fd5d8',
     '3a4eff55-5d01-48c8-a827-fca5732fd767',
     '3c5d59d2-6c3f-474b-95f4-ac6fcc192655',
+    '510f9741-823e-41a9-af2d-048895f990c0',
```
wjt, line 48: Please don't merge, we suddenly realised this data may be useful!

wjt: Rebased the branch and dropped this patch.

dbnicholson (approved):
My eyes glazed over on the `Uptime` commit, but it all seems reasonable.

PR #187, opened 2024-05-02
Ingest CPU flags from ComputerInformation event

wjt:
We would like to adjust the CPU information sent by eos-metrics-instrumentation to include CPU flags: statistics about how widely-supported certain extensions are will allow the freedesktop-sdk project to quantify the impact of (for example) assuming certain x86-64 extensions are present.

Previously, events with some given UUID are expected to have a payload of one specific GVariant type, and those types are typically not something extensible like `a{sv}`. This is quite annoying when we want to add a new field, particularly in this CPU data which just ends up being stuffed into a JSONB column in the database.

So, allow event models to specify a tuple of accepted GVariant types; then take advantage of this in the ComputerInformation model to understand a new type with an extra string of CPU flags.

https://phabricator.endlessm.com/T35351

- https://github.com/endlessm/eos-metrics-instrumentation/pull/113

### DIFF
--- a/azafea/event_processors/endless/metrics/v3/model/__init__.py
+++ b/azafea/event_processors/endless/metrics/v3/model/__init__.py
@@ -540,7 +540,10 @@ class ComputerInformation(SingularEvent):
     """
     __tablename__ = 'computer_information_v3'
     __event_uuid__ = '81f303aa-448d-443d-97f9-8d8a9169321c'
-    __payload_type__ = '(uuuua(sqd))'
+    __payload_type__ = (
+        '(uuuua(sqd))',
+        '(uuuua(sqds))',
+    )
 
     #: total RAM size in mebibytes (2^20)
     total_ram = Column(BigInteger, nullable=False)
@@ -550,20 +553,32 @@ class ComputerInformation(SingularEvent):
     used_disk = Column(BigInteger, nullable=False)
     #: free disk space in gibibytes (2^30)
     free_disk = Column(BigInteger, nullable=False)
-    #: array of CPU model (e.g. Intel(R) Core(TM) i7-5500U CPU @ 2.40GHz),
-    #: number of cores (e.g. 4) and maximum CPU speed in MHz or current CPU speed
-    #: if maximum can’t be determined (e.g. 3000.0)
+    #: array of objects containing the following keys:
+    #:
+    #: ``model`` (string)
+    #:     CPU model (e.g. ``Intel(R) Core(TM) i7-5500U CPU @ 2.40GHz``).
+    #: ``cores`` (integer)
+    #:     number of cores (e.g. ``4``).
+    #: ``max_frequency`` (float)
+    #:     maximum CPU speed in MHz, or current CPU speed if maximum couldn’t
+    #:     be determined (e.g. ``3000.0``).
+    #: ``cpu_flags`` (sorted array of unique strings)
+    #:     CPU flags (e.g. ``["de", "fpu", "vme", ...]``). Reported since
+    #:     Endless OS 6; this key is absent if the client did not send it.
     info = Column(JSONB, nullable=False)
 
     @staticmethod
     def _get_fields_from_payload(payload: GLib.Variant) -> Dict[str, Any]:
         cpu_info = []
         for item in payload.get_child_value(4).unpack():
-            cpu_info.append({
+            cpu = {
                 'model': item[0],
                 'cores': item[1],
                 'max_frequency': item[2],
-            })
+            }
+            if len(item) == 4:
+                cpu['flags'] = sorted(set(item[3].split()))
+            cpu_info.append(cpu)
         return {
             'total_ram': payload.get_child_value(0).get_uint32(),
             'total_disk': payload.get_child_value(1).get_uint32(),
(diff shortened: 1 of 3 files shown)


azafea/event_processors/endless/metrics/v3/model/__init__.py
```diff
@@ -540,7 +540,10 @@ class ComputerInformation(SingularEvent):
     """
     __tablename__ = 'computer_information_v3'
     __event_uuid__ = '81f303aa-448d-443d-97f9-8d8a9169321c'
-    __payload_type__ = '(uuuua(sqd))'
+    __payload_type__ = (
+        '(uuuua(sqd))',
+        '(uuuua(sqds))',
```
dbnicholson, line 545: So, if you're going to change the client, why not change this to an `a{sv}` like you talked about in the overview? I don't think you'd want to go so far as to just take it literally from the client and put it into the `info` column, but it seems like it would require no further payload type updates in the future.

One advantage there is that if we decide some field being sent by clients is no longer useful, we can just stop processing it here and don't have to worry that older clients keep sending it. On the other hand, you have to do a bit more type checking and/or conversion to ensure the stored JSONB data is in the format expected.

azafea/event_processors/endless/metrics/v3/model/_base.py
```diff
@@ -191,7 +191,7 @@ class MetricEvent(Base, metaclass=MetricMeta):
     __abstract__ = True
 
     __event_uuid__: str
-    __payload_type__: Optional[str]
+    __payload_type__: Optional[Union[str, Tuple[str, ...]]]
```
dbnicholson, line 194: This seems useful, but my interpretation is that you can't really have an event processor handle arbitrary payloads. It seems like the limitation here is that all the payload types must be able to be mapped to the same database schema.

In the past the way we would have handled this would be a `FooEvent2` with the new payload type, right?

Anyways, I think this makes sense. I can see how having the event processor handle multiple event UUIDs would be better in some ways, but I can see the utility of both should someone want to do the larger refactoring at some point. It's a lot of churn to add a new UUID when the processor can just handle different payloads appropriately. Since you're still type checking the payload regardless of how many types there may be, it seems to me that the multiple UUIDs per processor really only benefits the analysis side where you can munge multiple events into the same table. Is my understanding of this correct?

dbnicholson (approved):
I left some questions, but it seems fine as is. You certainly know more about this than I do.

azafea/event_processors/endless/metrics/v3/model/__init__.py
wjt: > So, if you're going to change the client, why not change this to an `a{sv}` like you talked about in the overview?

Inertia I guess!

> I don't think you'd want to go so far as to just take it literally from the client and put it into the `info` column, but it seems like it would require no further payload type updates in the future.
> 
> One advantage there is that if we decide some field being sent by clients is no longer useful, we can just stop processing it here and don't have to worry that older clients keep sending it. On the other hand, you have to do a bit more type checking and/or conversion to ensure the stored JSONB data is in the format expected.

You're probably right. I'll experiment with how each side looks with that change.

azafea/event_processors/endless/metrics/v3/model/_base.py
wjt: > It seems like the limitation here is that all the payload types must be able to be mapped to the same database schema.

Yep.

> In the past the way we would have handled this would be a `FooEvent2` with the new payload type, right?

Yes. Or: change the UUID, migrate the existing data, and discard any future submissions of data with the old UUID.

In this case it seems unnecessary to do any migration when the database schema already supports the change we want to make; and it's easy enough to keep receiving new data in the old format.

> Anyways, I think this makes sense. I can see how having the event processor handle multiple event UUIDs would be better in some ways, but I can see the utility of both should someone want to do the larger refactoring at some point. It's a lot of churn to add a new UUID when the processor can just handle different payloads appropriately. Since you're still type checking the payload regardless of how many types there may be, it seems to me that the multiple UUIDs per processor really only benefits the analysis side where you can munge multiple events into the same table. Is my understanding of this correct?

At present you have (in theory) the nice property that each event UUID always has a payload of a single known type, and if the payload type doesn't match you can probably just discard the event. The payloads are transmitted within a value of type `mv`, so you can always deserialize them with or without this property.

I say "in theory" because I have a dim memory that there were times in the past where we changed the type of an event's payload without changing its UUID. In one case I guess we changed an event from "no payload" to "has payload", which is why we have special handling of the `EmptyPayloadError` where if the UUID is in a `IGNORED_EMPTY_PAYLOAD_ERRORS` array we just discard the event.

Right now we (otherwise) don't discard events whose payload is of an unexpected type: they get stashed in the invalid_[singular|aggregate]_events_v3 tables, and we replay them every time we redeploy azafea. (Actually I just noticed that we never implemented that replay logic for aggregate events, whoops.)

azafea/event_processors/endless/metrics/v3/model/__init__.py
dbnicholson: Well, now that I looked at the client side changes, I would completely understand doing it this way for expediency. Changing it from a tuple to a dict might be more trouble than it's worth right now.

wjt: Looking at the submission side again… I think if I were to do something more invasive I would want to make some much bigger changes. For instance, the CPU info is an array, but that array always has exactly 1 element. I would want to send the CPU vendor, family & model. etc. etc.

So I will shelve the idea of further changes and undraft & land both sides of this.

PR #193, opened 2024-06-05
Update to Python 3.11

dbnicholson:
This supersedes #186 and is heavily inspired by https://github.com/endlessm/azafea-metrics-proxy/pull/27. Most of the changes were directly taken from that PR, but a few more changes were needed here since there's postgres usage and a slightly complex config setup for `queues`.

After this, only sqlalchemy isn't current (see #147) and it would be nice to drop types-redis but was a little more complex than I expected.


azafea/tests/test_cli.py
```diff
@@ -27,7 +27,7 @@ def process(*args, **kwargs):
 
         return process
 
-    config_file = make_config_file({'queues': {'some-queue': {'handler': 'azafea.tests.test_cli'}}})
+    config_file = make_config_file({'queues': {'some-queue': {'handler': 'azafea.tests'}}})
```
wjt, line 30: I don't understand this change.

> **tests: Make test queue handlers a package**
>
> Azafea is setup to expect the handler methods (`process` and
> `register_commands`) in a package with `migrations` as a sub-package.
> Otherwise newer alembic blows up when it tries to look for the
> `migrations` sub-package when it's the handler is defined as a module.

In this commit you're just changing these config files but you're not making any changes to `azafea.tests`. And that namespace doesn't define those two functions or the sub-package `migrations`.  How can this work / do anything?

azafea/tests/integration/migrations/test_migrations.py
```diff
@@ -7,26 +7,19 @@
...
-    @pytest.fixture(autouse=True)
-    def set_handler_module(self, handler_with_migrations):
-        self.handler_module = handler_with_migrations.basename
-        self._handler_path = handler_with_migrations
 
+class TestEmptyMigrations(IntegrationTest):
     def test_make_empty_migration(self):
         self.run_subcommand('make-migration', 'test_make_empty_migration')
 
-        migrations = (self._handler_path / 'migrations').listdir()
```
wjt, line 29: AFAICT `pathlib.Path` never had a `.listdir()` method – how did this work before?!

wjt (approved):
Hats off!

azafea/tests/test_cli.py
dbnicholson: I wrote that commit message hastily. What's happening is `get_callable` is being mocked just above that so that it magically returns a `process` method without having to add one to the actual package/module specified in `handler`. In essence, this is a fake handler and what's being specified doesn't matter except that it's an importable path. Most of the cases I think would be fine left as is, but I changed them all to be consistent since specifying a leaf module instead of a package would not be legitimate.

Where it actually matters is when alembic gets involved via `make-migration` or `migratedb`. In that case, we dynamically tell alembic that the migrations directories (as configured with `version_locations`) [are relative to the handler module](https://github.com/endlessm/azafea/blob/593f5fedb8d8d818534ef771296bf20a8202dfb0/azafea/migrations/utils.py#L26). If the "module" isn't a directory (i.e., a package), then alembic (or maybe our code driving alembic) fails.

azafea/tests/integration/migrations/test_migrations.py
dbnicholson: Because it used to be a [tmpdir fixture](https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmpdir-fixture), which returns a `py.path.local` object. I've never looked closely, but I think it's basically a precursor to `pathlib.Path`.

azafea/tests/test_cli.py
dbnicholson: If I'm being honest, I didn't really understand the issue completely until I wrote that comment.

wjt:
Looks good except for the documentation building test failing – will the same warning prevent publication on Rtd? Or we could merge & start testing the code changes in dev.

PR #195, opened 2024-06-06
docs/metabase: Null blank string values

dbnicholson:
Per the documentation[1][2], the parameters in the Table and Field update endpoints are "nullable value must be a non-blank string". Presumably earlier versions of Metabase converted blank strings to null internally. Make sure we use `None` if the stripped string value is empty.

1. https://www.metabase.com/docs/latest/api/table#put-apitableid
2. https://www.metabase.com/docs/latest/api/field#put-apifieldid

https://phabricator.endlessm.com/T32367

### DIFF
--- a/docs/source/metabase.py
+++ b/docs/source/metabase.py
@@ -8,7 +8,7 @@ import os
 from pathlib import Path
 import re
 import time
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, Union
 
 import sphinx.writers.text
 import requests
@@ -134,6 +134,14 @@ class MetabaseBuilder(DummyBuilder):
 
         return entry.get('data')
 
+    @staticmethod
+    def _none_if_blank(value: str) -> Union[str, None]:
+        """Return the value with whitespace stripped or None if it's blank."""
+        value = value.strip()
+        if len(value) == 0:
+            return None
+        return value
+
     def format_description(self, lines: List[str]) -> str:
         """Format tables and fields descriptions from ReST to plain text."""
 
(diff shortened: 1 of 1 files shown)


docs/source/metabase.py
```diff
@@ -134,6 +134,14 @@ def _read_session_data(self) -> Optional[Dict]:
 
         return entry.get('data')
 
+    @staticmethod
+    def _none_if_blank(value: str) -> Union[str, None]:
+        """Return the value with whitespace stripped or None if it's blank."""
```
dbnicholson, line 139: Suggestions welcome on this name. The name may be horrible regardless, but the awkward part to me is that it's returning the stripped value rather than just using that to determine if the value is blank.

wjt: I think this is fine, it's just a name. But here is an alternative with a more concise type, implementation, and slightly different name that directly mirrors the implementation:

```python3
@staticmethod
def stripped_or_none(value: str) -> Optional[str]:
    """Return the value with whitespace stripped or None if it's blank."""
    return value.strip() or None
```

wjt (approved):

docs/source/metabase.py
dbnicholson: That is much nicer. Maybe later.

Is `Optional` correct? I had that originally, but I thought it was more intended for arguments. However, it does just appear to be an alias for `Union[T, None]`.

wjt: I think it's just a bad name for that type.

Apparently these days you can spell Union[T, U] as `T | U` so this could become:

```python3
@staticmethod
def stripped_or_none(value: str) -> str | None:
    """Return the value with whitespace stripped or None if it's blank."""
    return value.strip() or None
```

PR #196, opened 2024-06-07
github: Use API key instead of username/password

dbnicholson:
An API key is the preferred way to authenticate to the Metabase API. The metabase builder supports API keys using the `METABASE_API_KEY` environment variable since 7337d7a.

https://phabricator.endlessm.com/T32367

This depends on https://github.com/endlessm/eos-administration/pull/1448.

### DIFF
--- a/.github/workflows/metabase.yml
+++ b/.github/workflows/metabase.yml
@@ -12,8 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     container: python:3.11-alpine
     env:
-      METABASE_USERNAME: ${{ secrets.METABASE_USERNAME }}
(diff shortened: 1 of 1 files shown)


dbnicholson:
The workflow is failing since I haven't actually deployed https://github.com/endlessm/eos-administration/pull/1448 yet. Maybe I'll just do that to test...

dbnicholson:
> The workflow is failing since I haven't actually deployed [endlessm/eos-administration#1448](https://github.com/endlessm/eos-administration/pull/1448) yet. Maybe I'll just do that to test...

Did that and now it works.

